PR #5, opened 2020-08-12
Remove CassandraLauncher

patriknw:
* use unique keyspace names in tests


patriknw:
> Should we add starting Cassandra via docker-compose?

Pls create an issue. Might need some discussion. Some might also promote TestContainers for tests, but based on the experience of that in Akka Projections I'm not overly enthusiastic.

PR #7, opened 2020-08-12
More about event sourcing

ennru:


ennru:
Applied review comments, I'd like to get it merged to evolve some structure.

Issue #6, opened 2020-08-12
Start Cassandra for running the services


ennru:
Add a docker-compose configuration for Cassandra to run the example easily in a local setting.

patriknw:
There is one we can take from https://github.com/akka/akka-persistence-cassandra/blob/master/docker-files/cassandra.yaml

Issue #18, opened 2020-08-14
Don't create keyspace and tables automatically


patriknw:
The `Main` of the sample is now creating keyspace and tables automatically, which is not our recommendation for production. Ok to do that from tests.

How should we deal with this for running locally? Easy of running vs the right thing for production?

patriknw:
@ennru suggested that this might be possible from docker compose

chbatey:
I don't think docker compose supports this, recommendations suggest editing the entrypoint/command of the image, which we don't want to do as we're using standard kafka/cassandra images.

We could provide a script that executes docker-compose and then waits for them to be up and then creates the schema. WDYT?

ennru:
> We could provide a script that executes docker-compose and then waits for them to be up and then creates the schema. 

SGTM. It should have the CQL in a separate file so we can reuse the script in other examples.

patriknw:
@chbatey killrvideo example has something https://github.com/KillrVideo/killrvideo-dse-config

chbatey:
That builds a new image that does it. I assume we want to use the standard
image

octonato:
A script will require the user to have `cqlsh` installed, no? 

We can avoid it with a minimal image packaging the scripts and starting as a side-cart to the cassandra container. We let docker-compose pick it as well. 

The side-cart can then create the schema. For the users there is only one thing to do, start the he docker compose.

octonato:
For clarify, I'm using the term side-cart, but I'm not suggesting a side-cart in k8s. Just a container in docker-compose to do some scripting and configuring the other containers.

patriknw:
I think `cqlsh` is available in the Cassandra with docker-compose, then the user can run the script with `docker run`.
Something similar to what is in https://doc.akka.io/docs/akka-projection/current/getting-started/running.html#running-the-projection

octonato:
I'm afraid this will be annoying to do repeatedly. 

First, the docker containers are prepended with the name of the current folder. And that also impact the name of the docker network to use. For instance, I just created a new seed in a folder called `demo`. 

I have the following containers:

```
demo_cassandra_1    Up About a minute   7000-7001/tcp, 7199/tcp, 9160/tcp, 0.0.0.0:9042->9042/tcp
demo_zookeeper_1    Up About a minute   22/tcp, 2888/tcp, 3888/tcp, 0.0.0.0:2181->2181/tcp
demo_kafka_1        Up About a minute   0.0.0.0:9092->9092/tcp
[...]

PR #35, opened 2020-08-19
Start Cassandra and Kafka via docker compose

ennru:
References #6

We might want to add creating the keyspace and tables, and the topics here, as well.


docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/application.conf
```diff
@@ -86,6 +86,6 @@ shopping-cart {
   askTimeout = 5 s
 
   kafka-topic = "shopping_cart_events"
-  kafka-bootstrap-servers = "localhost:9092"
+  kafka-bootstrap-servers = "localhost:9094"
```
patriknw, line 89: is it possible to keep this at 9092, because then it also works out of the box with a local Kafka installation without Docker?

there is also a corresponding config in the analytics service

patriknw:
> We might want to add creating the keyspace and tables, and the topics here, as well.

sounds good to create keyspace and tables here, added a note in #18 

Regarding the topic it seems to be created automatically anyway.

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/application.conf
ennru: Switched the ports around to expose Kafka on 9092.

patriknw: thanks

ennru:
There is something kafkaesque to the Kafka setup. That's for tomorrow.

patriknw:
I'll try this out, tomorrow

ennru:
I went a bit overboard it seems. This simpler config does the job.

patriknw (approved):
This works for me too. First time I ran the IntegrationSpec it failed with timeout on consuming from the topic, but after that it worked. Maybe the server wasn't ready.

ennru:
I believe the first run fails as the topic is not there yet when the consumer starts.

patriknw:
Topics seems to be created automatically, otherwise it would always fail. The topic name is unique for each test run.

PR #39, opened 2020-08-19
remove cluster roles

patriknw:
* simplify by not using roles
* they can be added in a separate "how to scale" step, see https://github.com/akka/akka-microservices-samples/issues/25


ennru (approved):
Better to spare this concept.
LGTM.

PR #41, opened 2020-08-19
Replaces Akka-HTTP with Akka-gRPC in the list of deps

ignasi35:
The list of dependencies list Akka-HTTP but the project really uses Akka-gRPC and Akka-HTTP is a transitive dep.


ennru (approved):

patriknw:
good change in docs, but note that http must be defined in the build since grpc has an older dependency and we want to use 10.2.0

ignasi35:
> good change in docs, but note that http must be defined in the build since grpc has an older dependency and we want to use 10.2.0

The overwrite makes a lot of sense. My suggestion in #40 was a bit in that line. the current writeup has a snippet with a huge list of dependencies. Explaining a bit more why each of those is necessary (or simply linking to the ref docs from a comment in the `build.sbt` would be helpful.

PR #44, opened 2020-08-20
tutorial outline

patriknw:
intended structure of the tutorial


ennru (approved):
Looks like a good plan.

PR #54, opened 2020-08-20
Shell script for scalafmt

ennru:


patriknw (commented):
Seems like unnecessary tooling to me. better to let intellij format on save, and for such small projects we could even enable format on compile in sbt

Issue #46, opened 2020-08-20
add readiness, liveness checks to the services 


ennru:
using Akka Management health checks

PR #64, opened 2020-08-21
Readiness

ennru:
I'm not sure how early we'd like to bring this in, but it makes sense immediatly after we have a service that persistes to Cassandra, IMO.

References #46

### DIFF
--- a/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
+++ b/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
@@ -19,3 +19,15 @@ akka {
   }
 }
 # end::cluster[]
+
+akka.management {
+  http {
+    hostname = "127.0.0.1"
+    port = 9101
+  }
+  health-checks {
+    readiness-checks {
+      akka-persistence-cassandra = "akka.persistence.cassandra.healthcheck.CassandraHealthCheck"
+    }
+  }
+}
--- a/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
+++ b/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
@@ -1,8 +1,6 @@
 # tag::cluster[]
 akka {
-  actor {
-    provider = cluster
-  }
+  actor.provider = cluster
 
   remote.artery {
     canonical.hostname = 127.0.0.1
@@ -10,11 +8,6 @@ akka {
   }
 
   cluster {
-    seed-nodes = [
-      "akka://Cart@127.0.0.1:2551",
-      "akka://Cart@127.0.0.1:2552"
-    ]
-
     downing-provider-class = "akka.cluster.sbr.SplitBrainResolverProvider"
   }
 }
--- a/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
+++ b/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
@@ -4,17 +4,15 @@ akka {
     provider = cluster
   }
 
-  # For the sample, just bind to loopback and do not allow access from the network
-  # the port is overridden by the logic in main class
   remote.artery {
-    canonical.port = 0
     canonical.hostname = 127.0.0.1
+    canonical.port = 2551
   }
 
   cluster {
     seed-nodes = [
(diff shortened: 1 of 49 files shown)


docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/node1.conf
```diff
@@ -0,0 +1,8 @@
+include "application"
```
patriknw, line 1: ok, yes better to manually override all ports than having it in code (Main)

Shall we name it `local1`, `local2` or `dev1`, `dev2` to make it clear that the purpose is for local development.

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
```diff
@@ -4,20 +4,30 @@ akka {
     provider = cluster
   }
 
-  # For the sample, just bind to loopback and do not allow access from the network
-  # the port is overridden by the logic in main class
   remote.artery {
-    canonical.port = 0
     canonical.hostname = 127.0.0.1
```
patriknw, line 8: Then we can move this to the other files also. Akka Bootstrap seems to like the default hostname anyway.

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
```diff
@@ -4,20 +4,30 @@ akka {
...
+    canonical.port = 2551
   }
 
   cluster {
     seed-nodes = [
```
patriknw, line 13: and seed-nodes config can also move to the other files, since we will use Bootstrap for the cloud

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
```diff
@@ -4,20 +4,30 @@ akka {
...
     ]
 
     downing-provider-class = "akka.cluster.sbr.SplitBrainResolverProvider"
   }
 }
 # end::cluster[]
+
+akka.management {
+  http {
+    hostname = "127.0.0.1"
```
patriknw, line 25: does it work without defining hostname?

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/node1.conf
```diff
@@ -0,0 +1,8 @@
...
+
+// node1 uses the default ports
+//akka.remote.artery.canonical.port =
+//akka.management.http.port =
+//shopping-cart.grpc.port =
```
patriknw, line 6: let's define those here anyway (repeat the defaults from application.conf)

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/node4.conf
```diff
@@ -0,0 +1,7 @@
+include "application"
+
+akka.remote.artery.canonical.port = 2554
+akka.management.http.port = 9054
+shopping-cart.grpc.port = 8054
```
patriknw, line 5: why different convention for these ports compared to node3?

docs-source/docs/modules/how-to/pages/index.adoc
```diff
@@ -0,0 +1,8 @@
+= How to ... with Akka
```
patriknw, line 1: good

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/cluster.conf
ennru: I'm afraid it doesn't

https://github.com/akka/akka-management/blob/9feaa68949e2ff7092a519d8595d8918f9584fd9/management/src/main/resources/reference.conf#L13

chbatey: It "might" work, depends on the default local interface.

It will work without this in k8s

patriknw: ok, let's revisit the hostnames in another PR. All 127.0.0.1 should be in local files (if they have to be defined).

patriknw (approved):
LGTM

PR #71, opened 2020-08-24
Make Antora site build from git sources

ennru:
This makes the `make` command work and it will produce the site from the current state of `main`.

`make html-author-mode` continues to show the latest local changes including all TODO and REVIEW sections.


patriknw (approved):
LGTM

Issue #27, opened 2020-08-17
setup PR validation


patriknw:
Travis and CLA checker

ennru:
CLA works now, as well.

patriknw:
@ennru but we are still not running tests?

ennru:
That's correct.

ennru:
All set up now.

PR #72, opened 2020-08-24
Run tests on Travis

ennru:
References #27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ script:
 
 jobs:
   include:
-    - stage: compile
+    - stage: test
       env:
         DIR="docs-source/docs/modules/cqrs/examples/shopping-analytics-service-scala"
         CMD="Test/compile; scalafmtCheckAll"
@@ -42,7 +42,7 @@ jobs:
       name: "Run Antora to create site"
 
 stages:
-  - name: compile
+  - name: test
     if: NOT tag =~ ^v
 
   - name: site
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 script:
   - cd "$DIR"
   - java -version
-  - if [ -f docker-compose.yml ]; then docker-compose ps; fi
+  - if [ -f docker-compose.yml ]; then docker-compose up -d; fi
   - sbt -jvm-opts .jvmopts-travis "$CMD"
 
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
     - stage: test
       env:
         DIR="docs-source/docs/modules/cqrs/examples/shopping-analytics-service-scala"
-        CMD="Test/compile; scalafmtCheckAll"
+        CMD="test; scalafmtCheckAll"
       name: "Shopping analytics service"
     - env:
         DIR="docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -28,7 +29,7 @@ jobs:
       name: "Shopping analytics service"
     - env:
         DIR="docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala"
-        CMD="Test/compile; scalafmtCheckAll"
+        CMD="test; scalafmtCheckAll"
```
patriknw, line 32: change stage to `test`?

patriknw:
Looking at the Travis log it seems to be a problem with the shutdown. Since it is starting to mark nodes as Leaving it must have passed the ordinary tests and the problem is in `afterAll`. "Failed to stop [IntegrationSpec] within [10 seconds]"

We could try to increase that timeout.

```
akka.actor.testkit.typed.system-shutdown-default
```

We can also enable DEBUG in logback-test.xml and we might see what part of the shutdown that is taking time. I suspect the close of the SendProducer.

.travis.yml
```diff
@@ -17,18 +17,19 @@ before_install:
...
+  - if [ -f docker-compose.yml ]; then docker-compose up -d; fi
   - sbt -jvm-opts .jvmopts-travis "$CMD"
 
 jobs:
   include:
-    - stage: compile
+    - stage: test
       env:
         DIR="docs-source/docs/modules/cqrs/examples/shopping-analytics-service-scala"
         CMD="Test/compile; scalafmtCheckAll"
```
patriknw, line 28: I think we can use "test" for all projects even though they (currently) don't have tests.

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/test/scala/sample/shoppingcart/IntegrationSpec.scala
```diff
@@ -128,10 +129,12 @@ class IntegrationSpec
     with Eventually {
   import IntegrationSpec.TestNodeFixture
 
+  private val logger = LoggerFactory.getLogger(classOf[IntegrationSpec])
+
   implicit private val patience: PatienceConfig =
     PatienceConfig(5.seconds, Span(100, org.scalatest.time.Millis))
 
-  private val grpcPorts = SocketUtil.temporaryServerAddresses(4, "127.0.0.1").map(_.getPort)
+  private val grpcPorts = SocketUtil.temporaryServerAddresses(3, "127.0.0.1").map(_.getPort)
```
patriknw, line 137: right, one was removed

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/test/scala/sample/shoppingcart/IntegrationSpec.scala
```diff
@@ -206,6 +210,11 @@ class IntegrationSpec
         event
       }
       .runForeach(kafkaTopicProbe.ref.tell)
+      .recover {
```
patriknw, line 213: `.failed.foreach`

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/test/resources/logback-test.xml
```diff
@@ -7,10 +7,11 @@
     </appender>
```
patriknw, line 7: we might have to add `[%X{akkaAddress}]` also, difficult to see from which node the logging is from otherwise

```
<pattern>[%date{ISO8601}] [%level] [%logger] [%X{akkaAddress}] [%marker] [%thread] - %msg%n</pattern>
```

PR #80, opened 2020-08-25
Renamings

patriknw:
Don't squash! Two separate renamings:

* rename packages, Refs #34
* rename cqrs to shopping-microservices, Refs #58

Also renamed the github repository to akka-platform-guide

PR #81, opened 2020-08-25
Use boostrap and initialize after cluster has formed

chbatey:
Two changes:

* Use bootstrap for local so it'll be the same for "production"
* Only start sharding/projections once the cluster has started to avoid some warnings from sharding

### DIFF
--- a/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
+++ b/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
@@ -1,9 +1,20 @@
 akka {
   remote.artery.canonical.port = 2551
-  cluster {
-    seed-nodes = [
-      "akka://Cart@127.0.0.1:2551",
-      "akka://Cart@127.0.0.1:2552"
+}
+
+akka.management.cluster.bootstrap.contact-point-discovery {
+  discovery-method = config
+  required-contact-point-nr = 1
+  # config service discovery never changes
+  stable-margin = 1 ms
+}
+
+akka.discovery.config.services {
+  // service name == ActorSystem name by default
+  "cart" {
+    endpoints = [
+      {host = "127.0.0.1", port = 9101},
+      {host = "127.0.0.1", port = 9102}
     ]
   }
 }
@@ -18,7 +29,7 @@ akka.discovery.config.services = {
   // The Kafka broker's bootstrap servers
   "shopping-kafka-broker" = {
     endpoints = [
-      { host = "localhost", port = 9092 }
+      {host = "localhost", port = 9092}
     ]
   }
 }
@@ -28,7 +39,7 @@ akka.discovery.config.services = {
   // The Shopping Order service's gRPC endpoint
   "order-service-grpc" = {
     endpoints = [
-      { host = "localhost", port = 8301 }
+      {host = "localhost", port = 8301}
     ]
   }
 }
--- a/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/scala/sample/shoppingcart/Main.scala
+++ b/docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/scala/sample/shoppingcart/Main.scala
@@ -1,13 +1,15 @@
 package sample.shoppingcart
 
-import akka.actor.typed.{ ActorSystem, Behavior }
-import akka.actor.typed.scaladsl.{ AbstractBehavior, ActorContext, Behaviors }
+import akka.actor.typed.{ActorSystem, Behavior}
+import akka.actor.typed.scaladsl.{AbstractBehavior, ActorContext, Behaviors}
+import akka.cluster.typed.{Cluster, SelfUp, Subscribe}
 import akka.grpc.GrpcClientSettings
+import akka.management.cluster.bootstrap.ClusterBootstrap
 import akka.management.scaladsl.AkkaManagement
 import akka.projection.cassandra.scaladsl.CassandraProjection
 import akka.stream.alpakka.cassandra.scaladsl.CassandraSessionRegistry
 import org.slf4j.LoggerFactory
-import sample.shoppingorder.proto.{ ShoppingOrderService, ShoppingOrderServiceClient }
+import sample.shoppingorder.proto.{ShoppingOrderService, ShoppingOrderServiceClient}
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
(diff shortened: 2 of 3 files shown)


docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
```diff
@@ -1,9 +1,20 @@
 akka {
   remote.artery.canonical.port = 2551
```
patriknw, line 2: then make it a one-liner

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
```diff
@@ -1,9 +1,20 @@
...
-    seed-nodes = [
-      "akka://Cart@127.0.0.1:2551",
-      "akka://Cart@127.0.0.1:2552"
+}
+
+akka.management.cluster.bootstrap.contact-point-discovery {
+  discovery-method = config
+  required-contact-point-nr = 1
+  # config service discovery never changes
+  stable-margin = 1 ms
```
patriknw, line 9: will that still join 9102 if 9101 is restarted?

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/scala/sample/shoppingcart/Main.scala
```diff
@@ -37,21 +39,22 @@ object Main {
...
-class Guardian(context: ActorContext[Nothing]) extends AbstractBehavior[Nothing](context) {
+class Guardian(context: ActorContext[SelfUp]) extends AbstractBehavior[SelfUp](context) {
   val system = context.system
-// end::start-grpc[]
+  // end::start-grpc[]
 
   startAkkaManagement()
 
-  ShoppingCart.init(system)
+  Cluster(system).subscriptions ! Subscribe[SelfUp](context.self, classOf[SelfUp])
```
patriknw, line 57: what warning is that fixing? would be nice to keep it simple if possible

patriknw (commented):
I like this

patriknw (commented):
Same changes in the other two services.
Let's merge https://github.com/akka/akka-platform-guide/pull/80 first.

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/scala/sample/shoppingcart/Main.scala
chbatey: With 1ms stable margin we don't get any, i changed this when i added bootstrap initially and we get lots of warnings before the cluster has started as the shard region  doesn't get a response from the coordinator as it hasn't startedyet

chbatey: I can remove but i think in general this is good practice. When in k8s and we have proper stable margin we'll get warnings if we start to use sharding before cluster has started.

Perhaps we need a simpler API for executing something when the cluster has formed for typed?

patriknw: I guess that is the "Probably, no seed-nodes configured and manual cluster or bootstrap join not performed"
Maybe we should look into those warnings in Akka then. Maybe it could log that at debug level until there are buffered messages?

chbatey: Yep as the readiness probe should stop any user messages going to the shard region

patriknw: > Yep as the readiness probe should stop any user messages going to the shard region

That was my thinking. Not a problem until there is traffic.

chbatey: I'll revert this and we can discuss in a different PR / make it nicer in Akka

chbatey: We'll also need to change sharded demons as they start sending messages which get buffered

docs-source/docs/modules/cqrs/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
chbatey: Yep it'll probe both the contact points and get seed nodes back from number 2. It'll only self join if no contact points return seed nodes

docs-source/docs/modules/shopping-microservices/examples/shopping-analytics-service-scala/build.sbt
```diff
@@ -1,8 +1,9 @@
-val AkkaVersion = "2.6.8"
+// FIXME once akka 2.6.9 is released
+val AkkaVersion = "2.6.8+71-57fb9e90"
 val AkkaPersistenceCassandraVersion = "1.0.1"
 val AlpakkaKafkaVersion = "2.0.4"
 val AkkaHttpVersion = "10.2.0"
-val AkkaManagementVersion = "1.0.8"
+val AkkaManagementVersion = "1.0.8+35-9feaa689+20200825-1429"
```
chbatey, line 6: This contains an additional config which allows us to disable all contacts being up before starting a clustser. Useful for us to include multiple contacts for bootstrap but it still work if the user only starts 1

chbatey: https://github.com/akka/akka-management/pull/749

docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/build.sbt
```diff
@@ -1,8 +1,10 @@
-val AkkaVersion = "2.6.8"
+// FIXME once akka 2.6.9 is released
+val AkkaVersion = "2.6.8+71-57fb9e90"
```
chbatey, line 2: Sick of that sharding log. I guess we'll make changes to Akka as we go and using a snapshot will be useful

patriknw (approved):
LGTM

patriknw:
Compilation error in IntegrationSpec. Would be good if you run that also, since Travis testing isn't working yet.

PR #83, opened 2020-08-25
gRPC server part as separate example project

patriknw:
* 00-shopping-cart-service-scala - empty project, starting point
* 01-shopping-cart-service-scala - Part 3: First gRPC service


docs-source/docs/modules/shopping-microservices/examples/01-shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
```
patriknw, line 1: added in this step

docs-source/docs/modules/shopping-microservices/examples/01-shopping-cart-service-scala/src/main/scala/shopping/cart/Main.scala
```diff
@@ -0,0 +1,40 @@
...
+}
+
+class Guardian(context: ActorContext[Nothing]) extends AbstractBehavior[Nothing](context) {
+  val system = context.system
+
+  startAkkaManagement()
+
+  val grpcInterface = system.settings.config.getString("shopping-cart.grpc.interface")
+  val grpcPort = system.settings.config.getInt("shopping-cart.grpc.port")
+  ShoppingCartServer.start(grpcInterface, grpcPort, system)
```
patriknw, line 31: updated in this step

docs-source/docs/modules/shopping-microservices/examples/01-shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServer.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.util.Success
+
+import akka.actor.typed.ActorSystem
+import akka.grpc.scaladsl.ServerReflection
+import akka.grpc.scaladsl.ServiceHandler
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model.HttpRequest
+import akka.http.scaladsl.model.HttpResponse
+
+object ShoppingCartServer {
```
patriknw, line 16: added in this step

docs-source/docs/modules/shopping-microservices/examples/01-shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala
```diff
@@ -0,0 +1,16 @@
+package shopping.cart
+
+import scala.concurrent.Future
+
+import org.slf4j.LoggerFactory
+
+class ShoppingCartServiceImpl extends proto.ShoppingCartService {
```
patriknw, line 7: added in this step

patriknw:
I don't know if it's too early to break up the code in steps like this? I could remove all files that are the same as the full sample and only keep the changed.

ennru:
It is a bit early, but maybe worth the extra work so that we can get some of the structure we'd like.
The final example will still change a bit and updating all these clones might become a bit messy.

patriknw:
I'll only include the changed files, and to begin with I think we will get away with 3 copies for the early steps and then full sample. Propagating such changes with Intellij diff tool is ok.

patriknw:
Removed files that are the same as in full sample.

patriknw:
I'll update this with latest from master before merging

Issue #79, opened 2020-08-25
Use of bootstrap and discovery


chbatey:
Bootstrap - should we use it locally too so we don't end up having to describe seed nodes and bootstrap. Also avoids not having bootstrap.start being conditional for local.

Discovery - consider not using it for Kafka/gRPC/Cassandra. In k8s these can be headless services and for local we can override them to be localhost with config.

patriknw:
Bootstrap was added in https://github.com/akka/akka-platform-guide/pull/81

PR #92, opened 2020-08-26
Don't use disovery for kafka and gRPC

chbatey:
I don't see it adds any value. In kubernetes we'll likely access
dependencies via a service so a simple config property is fine.

If we agree i'll change the other services too.


docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
```diff
@@ -26,24 +26,10 @@ akka.remote.artery.canonical.hostname = "127.0.0.1"
...
-akka.discovery.config.services = {
-  // The Shopping Order service's gRPC endpoint
-  "order-service-grpc" = {
-    endpoints = [
-      {host = "localhost", port = 8301}
-    ]
-  }
-}
+order-service.host = "localhost"
+order-service.port = 8301
```
patriknw, line 34: also put these in application.conf, even though they will not be used until we deploy

docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
```diff
@@ -26,24 +26,10 @@ akka.remote.artery.canonical.hostname = "127.0.0.1"
...
-
-akka.discovery.config.services = {
-  // The Kafka broker's bootstrap servers
-  "shopping-kafka-broker" = {
-    endpoints = [
-      {host = "localhost", port = 9092}
-    ]
-  }
-}
+shopping-cart.kafka.bootstrap-servers = "localhost:9092"
```
patriknw, line 29: also in kafka.conf

patriknw (commented):
looking good, seems easier to me

chbatey:
I'll do this for all the services now

patriknw:
sorry, I also removed that file

docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/local-shared.conf
patriknw: I have changed those in other places to be full names: shopping-order-service

patriknw (approved):
LGTM, after small rename

patriknw:
Refs #79

Issue #59, opened 2020-08-20
[EPIC] Concepts around reactive microservices


ennru:
Some material from https://www.lagomframework.com/documentation/1.6.x/scala/CoreConcepts.html would be good to include here.

ignasi35:
From https://www.lightbend.com/ebooks/reactive-microservices-architecture-design-principles-for-distributed-systems-oreilly (via https://docs.google.com/presentation/d/14t0bgwdcERTtCgp2sDIWeGzFG2KdBVqgOOw_S2mqF1U/edit#slide=id.g3766cd14f5_0_47) we have a rather large list of concepts we could cover when talking about microservices: 

(I did a quick-and-dirty ordering/grouping)


--------------------------------

Isolation

Bulkheading (Failure isolation)

Autonomy


Decoupling in time and space
Services that are mobile

Microliths
Eventual consistency
Causal consistency

Conway’s Law

--------------------------------

Amdahl’s law
Universal Scalability Law

--------------------------------

(General)
Functional programming / languages
Finite-state machine
Idempotency (of side-effects)
Streaming
Leaky abstractions
Conflict-free Replicated Data Types (CRDTs)


(General - ES/CQRS)
Command Query Responsibility Segregation (CQRS)
Event sourcing
Event storming
Single Writer Principle

The Saga pattern
Compensating actions
Memory image pattern

Consistency guarantees and at-least once delivery
Polyglot persistence

(General - Big Data)
Lambda architecture
Streaming and data in motion


(General - SW practices)
Single Responsibility Principle (from SOLID)
Postel’s Law / Robustness Principle
Own state exclusively

--------------------------------

(actors)
Supervision
Location transparency
Dead letter queue

Backpressure
Circuit breakers

Reactive Streams
Reactive Programming
Reactive Systems

--------------------------------

(DDD)
Bounded Context, which is part of DDD
Stateful aggregate
Consistency boundaries
Events-First Domain-Driven Design (therefore assumes basic knowledge of DDD too)
Aggregates - Unit of consistency
Anti-Corruption Layer
Events and Commands

--------------------------------

(deployment concerns)
Service registry - Client-side service discovery, Server-side service discovery

(deployment tech)
AWS Lambda
Docker
Kubernetes


--------------------------------


(other)
ACID 2.0
Event-driven model using Pub/Sub over a message broker (like Kafka)
Asynchronous message-passing
Active-Passive state replication
Virtual addressing and stable references
Dynamic composition
[...]

PR #94, opened 2020-08-26
Example usage of ebook include for isolation and autonomy

ignasi35:
References #59 

These look like: 

Screenshot 2020-08-26 at 11 49 40

and

Screenshot 2020-08-26 at 11 49 46


patriknw (approved):
LGTM

Issue #77, opened 2020-08-25
grpc network interface conf


patriknw:
The config of the grpc interface is still defined as 127.0.0.1. We need different config for application.conf and local.conf

Two places:
```
shopping-cart.grpc.interface
shopping-order.grpc.interface
```

PR #100, opened 2020-08-27
Set grpc interface to localhost in local-shared

chbatey:


patriknw:
Refs https://github.com/akka/akka-platform-guide/issues/77

patriknw (approved):
LGTM, that could be a good default to start with. We might want to revisit when we deploy to real. Maybe a security concern to bind to 0.0.0.0? Maybe we will use some env variable?

PR #108, opened 2020-08-27
config consistency

patriknw:


docs-source/docs/modules/shopping-microservices/examples/shopping-analytics-service-scala/src/main/resources/kafka.conf
```diff
@@ -1,4 +1,4 @@
-shopping-analytics {
+shopping-analytics-service {
```
patriknw, line 1: this "full" naming will make it easier to generate stuff in the seed template

ignasi35 (approved):
LGTM. Very in favor of using more unique, verbose names for each elemt in the application (actor system, public addressable name, etc..). We've made the mistake in the past of calling everything "shopping-cart" and it's very hard to track which String is related to which.

Issue #91, opened 2020-08-26
Tutorial navigation


patriknw:
I don't know if we will get navigation for free with the real style. Otherwise we need to add links to next step at the the end of the tutorial pages.

patriknw:
Let's add this. We can have 'Learn more' section at the end of each with link to next (and prev) and a few related links to reference docs

patriknw:
@ignasi35 said:

> Having automatic Prev/Next would be great. Adding a custom admonition (https://asciidoctor.org/docs/user-manual/#admonition) for related topics would be ideal.

ignasi35:
I don't know much about custom admonitions but maybe we can use a [custom icon](https://github.com/asciidoctor/asciidoctor-pdf/blob/master/docs/theming-guide.adoc#key-prefix-admonition-icon) using the [fa-books](https://fontawesome.com/icons/books?style=duotone). 

`📚 Learn  more: ...`

patriknw:
I'll add something, and we can revisit the stylish solution later.

ignasi35:
for the record: https://gitlab.com/antora/antora-ui-default/-/blob/master/src/partials/pagination.hbs

PR #123, opened 2020-08-28
Add Learn more sections

patriknw:
* with links to next/prev step

References #91


ignasi35 (approved):
LGTM. Good to have something while we don't solve the automation

Issue #111, opened 2020-08-27
Add canonical URL 


ignasi35:
In #85 we added content copy/pasted from https://developer.lightbend.com/docs/introduction/traditional-architecture.html. The page https://github.com/akka/akka-platform-guide/blame/main/docs-source/docs/modules/concepts/pages/concepts-traditional-architecture.adoc should include a `canonical` tag pointing to the page in `https://developer.lightbend.com...`.

patriknw:
Does that mean that if ppl google it they will end up at the developer.lightbend page?

ignasi35:
> Does that mean that if ppl google it they will end up at the developer.lightbend page?

Ah, ofc! I guess we should consider adding the canonical in `https://developer.lightbend.com` to point to `https://guide.akka.io/...` (that is, _here_).

ignasi35:
@jtownley, we probably want to review https://developer.lightbend.com/docs/introduction/planning/microservices-project.html and https://developer.lightbend.com/docs/introduction/traditional-architecture.html to edit/remove(?) them and promote the contents _here_. WDYT?

patriknw:
Thanks. I think we will get back to this when we know more concretely what the overlap will be.

ignasi35:
> I think this falls into the "How" side of the process and I am very happy with the execution of these types of decisions being development ones.
> If your looking for general opinion I think having 1 place only for documentation would be good. If its cross project developer.lightbend.com seems good.
> If you are concerned that there is a business driver here that I might be missing let me know.

@jtownley the problem with https://developer.lightbend.com/docs/introduction/ is that it contains docs for multiple products across the Lightbend Stack (some of which fall out of the Akka Platform Guide scope). IMHO, we should move anything that's specific to the Akka Platform to the new Guide we're putting together, but doing that will make the contents of https://developer.lightbend.com/docs/introduction very scarce. My main concern is finding the owner for that set of docs(https://developer.lightbend.com/docs/introduction/) so we can decide a plan for it.

patriknw:
The guide will be published at developer.lightbend.com. Let's address any overlap between the introduction and the guide in another ticket (updating the introduction).

PR #129, opened 2020-08-28
[Concepts] Introduces circuit breakers

ignasi35:
References #59

This is a copy/paste version of https://doc.akka.io/docs/akka/current/common/circuitbreaker.html .
We will need to setup a canonical URL (#111) in one of the two pages.


ignasi35:
> I'm not sure what's the best approach here. Maybe linking to the Akka docs would be good enough right now?

@jtownley here's another example of how introducing the Akka Platform guide will impact the location of already-existing documentation (see discussion on #111). This case, the existing content is part of doc.akka.io and is provided by the `akka-docs/` project in https://github.com/akka/akka/. 

My point is, we will need to approach the `Concepts` (and maybe other sections) of the guide with a higher-level thinking.

patriknw:
> I'm not sure what's the best approach here. Maybe linking to the Akka docs would be good enough right now?

I'm not sure I understand what the problem with including the proposed content as a concept here even though it's the same introduction in the Akka reference documentation. It's a high level description.

The reference docs could be changed and instead link to this concept page, but for the reader's convenience it's good with an intro like that also in reference documentation without having to jump to another place.

docs-source/docs/modules/concepts/pages/circuit-breakers.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+    - Callbacks can be used eg. to collect statistics about all invocations or to react on specific call results like success, failures or timeouts.
+    - Supported callbacks are: *onCallSuccess*, *onCallFailure*, *onCallTimeout*, *onCallBreakerOpen*.
+    - These are executed in the `ExecutionContext` provided.
+   
+The following diagram illustrates the Finite State Machine described above:
+[#circuit-breaker-states]
+.States of a Circuit Breaker
+[caption=""]
+image::concepts/circuit-breaker-states.png[States of a Circuit Breaker]
```
ennru, line 70: Add links to the reference docs?

ignasi35:
I just realised we introduce the 'Circuit Breakers' concept but it is never mentioned or used in the tutorial. This is a smell.

IIUC, of all client code used in the sample app for the tutorial, only the akka-persistence-cassandra code has an explicit circuit breaker that the user may tune.

patriknw:
> Circuit Breakers concept but it is never mentioned or used in the tutorial.

I think that is more of an advanced topic that would fit into a how-to. I created issue https://github.com/akka/akka-platform-guide/issues/143 for that.

ennru (approved):
LGTM.

PR #131, opened 2020-08-31
Separate folders for separate sections

ignasi35:
References #59

More reorganizing. This PR does two things: 

1. puts concepts pages in separate folders to more easily navigate files
2. Adds `TODO` in pages that have pending content


docs-source/docs/modules/concepts/pages/index.adoc
```diff
@@ -14,18 +14,18 @@ Reactive Microsystems, and a high level overview of the key concepts that power
 
 == Properties of Reactive Microsystems
 
-* xref:isolation.adoc[]
-* xref:autonomy.adoc[]
-* xref:mobility.adoc[]
+* xref:properties/isolation.adoc[]
```
patriknw, line 17: While it is nice with a structure it can become more difficult to refactor the structure later if these become public URLs. Do you think we will have so many concept pages that it is a problem to have them all in a flat physical location?

docs-source/docs/modules/concepts/pages/index.adoc
```diff
@@ -57,11 +57,11 @@ overview of some features in the Akka toolkit you are likely to use:
...
-* xref:services.adoc[]
-** xref:service-discovery.adoc[]
-** xref:service-clients.adoc[]
-** xref:grpc-services.adoc[]
-* xref:internal-and-external-communication.adoc[]
+* xref:deployment/services.adoc[]
+** xref:deployment/service-discovery.adoc[]
+** xref:deployment/service-clients.adoc[]
+** xref:deployment/grpc-services.adoc[]
+* xref:deployment/internal-and-external-communication.adoc[]
```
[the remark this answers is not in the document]
ignasi35: The organizing per folder is sync'd with the categorization in the index page of the concepts module.

All those pages are WIP (include a TODO) and for the time being I just had them moved there for coherence (mimc the categorization of the index page).

docs-source/docs/modules/concepts/pages/index.adoc
ignasi35: Yeah, I was not impressed by the lack of refactoring features in the Idea plugin for asciidoc. I hoped a file move would manage all references and that's not the case.

I'm divided: I tend to prefer deep folder structures with very few files to better visualize what's related to what, but in this case I'm not sure we're gaining anything.

ignasi35: I'll keep this PR as DRAFT for reference as we discuss the final organization of these files.

patriknw: I'm not talking about problems for us to move the files, but the public URLs once we have published the docs. Having to add redirects because structure is changed (or renamed).

ignasi35: I wasn't thinking about the public URLs.

For the record (slightly related): https://docs.antora.org/antora/2.3/page/page-aliases/#page-aliases-attribute

ignasi35: TBH, I'd love if I could clearly declare the mapping of public URLs to internal files so the file/folder names didn't leak.

patriknw:
I would prefer that we keep it simple and use a flat structure to start with. We can group files together later when the structure has settled.

ignasi35:
We may eventually revisit the idea of keeping fewer files per folder but it's early and we're still moving content around too much.

PR #135, opened 2020-09-01
Adding review and todo comments

rstento:
This is not necessarily a PR to merge, although feel free if you like. It contains some initial thoughts about the content and organization as well as some asciidoc markup tips. You'll of course be able to see them in the diff, but to see them in the HTML, you will need to build using make html-author-mode. I put most in ifdef::review blocks.


patriknw:
Thanks for comments so far. Did you also see the tutorial (example) section, which is where we have spent most time so far:
docs-source/docs/modules/shopping-microservices/pages/

rstento:
I didn't because I know it will take the most time. I think it's important
as we discussed in the user experience group to have a particular audience
in mind and understand what their learning goals are. Then, I'll be glad to
run through it myself as a tester of the lowest common denominator :-)

Ruth Stento
User Experience Lead
Lightbend

512-696-1448
<http://lightbend.com>

ennru (approved):
Great input, I'll merge it and work in those directions.

PR #148, opened 2020-09-03
Adapt sample app build structure to CMT

eloots:
- Adds two steps as an example with a minute difference in a README
  file to illustrate the principle

- Assuming the project is cloned in a folder named `/AkkaSource/akka-platform-guide`,
  we can use the tooling to create additional steps, generate a linearized
  version for showing the diffs between different steps.

NOTE: One setting that wasn't applied in the merged project is the
      `Global / cancelable := false // ctrl-c` setting. It would be
      trivial to change the tooling to add this automatically.

Example CMT commands

- Regenerate the global build.sbt file:

```
cmt-mainadm -b -dots \
    -cfg docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/course-management.conf \
    /AkkaSource/akka-platform-guide
```

- Generate a 'minimised' linearized version of the master:

```
cmt-linearize -f -m -dot \
    -cfg docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/course-management.conf \
    /AkkaSource/akka-platform-guide \
    /Users/ericloots/tmp/lin
```

This will create a folder `/Users/ericloots/tmp/lin/akka-platform-guide`
holding the linearized repository

More info in the documentation which you can find here:

  https://eloots.github.io/course-management-tools/

References #xxxx


patriknw:
Thanks for preparing this @eloots .
Making the tutorial a full course is not in our scope. The example projects should be clean plain examples without any CMT traces.

What we might find useful is cmt-linearize and cmt-delinearize for maintaining changes to the steps of the tutorial. Is that possible without adding CMT to the build?

I tried by creating step directories as copies of shopping-cart-service-scala and then:
```
cmt-linearize --bare-lin-repo docs-source/docs/modules/shopping-microservices//examples/shopping-cart-service-scala ~/tmp/lin
```

but it complained about YOU HAVE CHANGES IN YOUR GIT WORKSPACE
I have everything committed and even tried after `git clean -fxd`

eloots:
Hi @patriknw Sorry for the delayed reaction, but I'm on vacation and I didn't notice your comment until now.

Not sure what you mean with _"without any CMT traces"_. What I can say is that, in order to utilise _linearize_ and _delinearize_, you need to have a build structure that meets some minimal criteria. These criteria are met by the changes I applied to the sbt build definition. As a result, the project is turned into a multi-project build with one project per solution step and a root build file that is generated by the tooling.

I applied some changes to the tooling as to be able to generate a minimalistic linearized version; the latter is still a multi project build with a root project and a 'solution' project.

The conversion between the master project and the linearized version (and back) uses git (as each step in the 'solution' is a commit in the linearized [git] repository). As a consequence, the process of linearisation requires you git workspace to be clean (see this section of the docs [here](https://eloots.github.io/course-management-tools/docs/reference-linearize#linearizing-a-cmt-main-repository)).

This means that in your workflow, you should add any changes in your workspace to the staging area and commit anything in the latter before linearizing the master repo. (In practice, that is not a big deal, because you can always do a soft reset to HEAD~ should you wish to do so).

The `build.sbt` file of the linearized project looks as follows:

```
Global / onChangedBuildSource := ReloadOnSourceChanges

Global / onLoad := {
  (Global / onLoad).value andThen (state => "project steps" :: state)
}

lazy val `steps` = (project in file("steps"))
  .settings(ThisBuild / scalaVersion := Version.scalaVersion)
  .configure(CommonSettings.configure)
  .enablePlugins(AkkaGrpcPlugin)`
```

patriknw:
> Not sure what you mean with "without any CMT traces". What I can say is that, in order to utilise linearize and delinearize, you need to have a build structure that meets some minimal criteria.

The projects for each step should be self contained with its own build file. Those build files should not have any dependencies ("traces") of CMT. My naive misunderstanding was that the linearize tool wouldn't care about the sbt builds but only look at the directory structure and git.

Thanks again for preparing this, but it seems like we don't want to add CMT to this tutorial. At least not now. We could revisit that later but then I think the motivation would be that it becomes part of some training material.

> the process of linearisation requires you git workspace to be clean

It's as clean as it can be. I even tried from a new clone and after `git clean -fxd`.

eloots:
> The projects for each step should be self contained with its own build file. Those build files should not have any dependencies ("traces") of CMT. My naive misunderstanding was that the linearize tool wouldn't care about the sbt builds but only look at the directory structure and git.

It depends on what your definition is of "traces" of CMT. The only two lines in the build definition that can be considered as such are:

```
Global / onChangedBuildSource := ReloadOnSourceChanges

Global / onLoad := {
  (Global / onLoad).value andThen (state => "project steps" :: state)
}
```

The first one can be removed by specifying the `-nar` option when running `linearize`. The second one 'positions' sbt on the [in this case] `steps` project. Because the use case for you is to be able to shows the git diffs between different steps, that line can be left out too.

Also, there is a `course-management.conf` file in the root folder of the example. That file can be anywhere in the folder structure of the overall git project. For example, the file could be move on level up and named `akka-platform-guide.cfg`

> It's as clean as it can be. I even tried from a new clone and after git clean -fxd.

There must be something else going on here that explains this. I suspect a mistake in the folder locations (of the master) or in the configuration file. I propose to have a short call as it should be easy to sort out what the problem is.

patriknw:
> It depends on what your definition is of "traces" of CMT

but it also requires them to a multi project build instead of a self contained build.sbt in each step project?

eloots:
> but it also requires them to a multi project build instead of a self contained build.sbt in each step project?

It does in the master, but that’s the essence of being able to retain 2 dimensions of changes:
- changes between “steps”
- overall history of all code versions

The first is maintained via the multi-project aspect of the build, the second via git history.

Also worth noting is that this approach makes certain tasks that occur rather regularly trivial to execute. Think about updating versions of dependencies. Also think about CI/CD that can become simpler as testing each step in the master repo is trivial whereas doing this on a repo in which each step is a commit requires some custom approach. Examples of that can be found in the CI/CD setup of CMT itself: at present, that repo contains 5 template CMT projects that are used to test the tools automatically on PR submits. 

Coming back to the use case this platform focuses on namely, having a git repo that allows one to inspect the diffs between steps, the build def for the linearized repo could be a “normal” single project build and I can adapt the tooling to support this.

The only condition is that all build “customization” should be inserted in the root build.sbt file via `Commonsettings.settings` (and, depending on the use case `Commonsettings.configure`.

PR #159, opened 2020-09-04
Kafka config use in the tutorial (no Service Discovery used)

ignasi35:

### DIFF
--- a/docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/kafka.conf
+++ b/docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/kafka.conf
@@ -2,8 +2,6 @@ shopping-cart-service {
 
   kafka.topic = "shopping_cart_events"
 
-  kafka.producer = ${akka.kafka.producer}
-
   # update to a service once k8s deployment is done
   kafka.bootstrap-servers = "localhost:9092"
 }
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/shopping-microservices/examples/shopping-cart-service-scala/src/main/resources/kafka.conf
```diff
@@ -2,9 +2,7 @@ shopping-cart-service {
 
   kafka.topic = "shopping_cart_events"
 
-  kafka.producer: ${akka.kafka.producer} {
-    service-name = "shopping-kafka-broker"
-  }
+  kafka.producer: ${akka.kafka.producer}
```
patriknw, line 5: Without the `{` I find that syntax strange. Include empty `{}` or
```
kafka.producer = ${akka.kafka.producer}
```

ignasi35: You're right, I missed the `:`. Using the `=` is the cleanest.

ennru (commented):
Alternatively, don't introduce `kafka.producer` and just use the default `akka.kafka.producer` from `ProducerSettings.ConfigPath` instead.

patriknw:
> Alternatively, don't introduce kafka.producer

Right, that would only be needed if having to connect to two or more Kafka brokers?

ennru:
> Right, that would only be needed if having to connect to two or more Kafka brokers?

Only if you want to pass different settings from `application.conf` to more than one producer. The broker addresses are most often passed in code (`withBootstrapServers`).

patriknw:
ok, [those settings](https://github.com/akka/alpakka-kafka/blob/master/core/src/main/resources/reference.conf#L6) look rather advanced and there is no obvious reason why different settings would be needed for different producers. Let's simplify and use `akka.kafka.producer` directly instead.

patriknw (approved):
LGTM, nice

ignasi35:
Moving out of the complexity and into defaults made it all cleaner.

ennru (approved):

PR #151, opened 2020-09-03
Howto scale independently

ignasi35:
References #25

PR #161, opened 2020-09-07
Inline unnecessary val (amends PR #151)

ignasi35:
References https://github.com/akka/akka-platform-guide/pull/151#discussion_r484405039


patriknw (approved):
LGTM, thanks

PR #155, opened 2020-09-04
Howto  reduce the build dependencies

ignasi35:
References #51

PR #164, opened 2020-09-09
Introduce concept 'Consumer-only services'

### LINKED ISSUE #154: [Concept or HowTo] Consumer-only services
There's a common pattern in microservices where a microservice doesn't expose any blocking endpoint (HTTP/gRPC) and, instead only consumes broker topics and emits data into other brokers or storages.

Refs: https://github.com/lagom/lagom/issues/1288
Refs: https://www.lagomframework.com/documentation/current/scala/KafkaClient.html#Subscriber-only-Services

ignasi35:
Fixes #154 
Needed in #155 (merge that and this and I'll amend later)

### DIFF
--- a/docs-source/docs/modules/concepts/pages/consumer-only-services.adoc
+++ b/docs-source/docs/modules/concepts/pages/consumer-only-services.adoc
@@ -1,0 +1,14 @@
+= Consumer-only Services
+:toc:
+:toc-title: ON THIS PAGE
+:toclevels: 2
+
+include::partial$include.adoc[]
+
+In architectures based on microservices, the most common structure of a microservice includes:
+
+. a blocking API (REST, gRPC, HTTP)
+. interaction with a broker (emit or consume messages)
+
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/concepts/pages/consumer-only-services.adoc
```diff
@@ -0,0 +1,14 @@
+= Consumer-only Services
+:toc:
+:toc-title: ON THIS PAGE
+:toclevels: 2
+
+include::partial$include.adoc[]
+
+In architectures based on microservices, the most common structure of a microservice includes:
+
+. a blocking API (REST, gRPC, HTTP)
```
ennru, line 10:
```suggestion
. a synchronous API (REST, gRPC, HTTP)
```

ennru (commented):
I'm not sure this is a complete text it doesn't explain what it title indicates.

patriknw:
I'm not convinced this concept is of much value. To me it's obvious that there can be Consumer-only services. It's just a subset of the optional pieces that we say that an Akka Microservice typically consists of.
There could also be Producer-only service, gRPC-only, ...

ignasi35:
Yeah, I guess we can leave the mention I added in #155 as is and move on.

Issue #127, opened 2020-08-28
Remove sbt dependencies from index page


patriknw:
The exact dependencies are not important as the very first thing to show.

The high level list is good but can be updated with Alpakka Kafka and Bootstrap.

PR #178, opened 2020-09-10
remove dependencies from tutorial index.adoc, #127

patriknw:
References #127


ignasi35 (commented):
Having removed this, there's no place in the whole tutorial presenting any sbt code. Maybe we can add one file `appendix` page for the build tools code?

patriknw:
Explaining how to use sbt is completely out of scope for this guide. I don't know if the build needs explanation? "should just work". Anyway, not essential part of tutorial.

Issue #179, opened 2020-09-10
Structure of the concepts


patriknw:
Request for input from Ruth when she reviews the guide in more depth.

Initially she said

> we might want to reconsider renaming "Concepts" and breaking it up into top level titles that might be more meaningful to readers, such as "Why Reactive", "What's Reactive" -- just brainstorming here.

ignasi35:
@rstento we've been discussing some option in #184

patriknw:
Was this automatically closed by the merge? I don't think we are completely done here yet.

ignasi35:
> Was this automatically closed by the merge? I don't think we are completely done here yet.

Agreed. IDK what happened.

PR #184, opened 2020-09-10
Renames concepts sections

ignasi35:
An idea to tackle #179

cc @rstento

### DIFF
--- a/docs-source/docs/modules/concepts/nav.adoc
+++ b/docs-source/docs/modules/concepts/nav.adoc
@@ -1,5 +1,5 @@
 .xref:index.adoc[Concepts]
-* Properties of Reactive
+* What is Reactive?
 ** xref:isolation.adoc[]
 ** xref:autonomy.adoc[]
 ** xref:mobility.adoc[]
@@ -7,7 +7,7 @@
 ** xref:message-driven-event-driven.adoc[Event-Driven]
 ** xref:state-ownership.adoc[]
 
-* Patterns and techniques for Reactive
+* How to design Reactive?
 ** xref:slicing-the-monolith.adoc[]
 ** xref:event-sourcing.adoc[]
 ** xref:cqrs.adoc[]
@@ -19,8 +19,9 @@
 ** Reactive Streams
 *** Backpressure
 ** xref:ddd.adoc[Brief Summary of DDD concepts]
+** xref:internal-and-external-communication.adoc[]
 
-* Tools for Reactive
+* What is in Akka for Reactive?
 ** Actor model
 *** Supervision
 ** Akka Streams
@@ -29,17 +30,16 @@
 *** SBR
 ** Akka Persistence
 ** Akka Projections
-
-* Deploying Reactive
-** xref:services.adoc[]
-*** xref:service-discovery.adoc[]
-*** xref:service-clients.adoc[]
-*** xref:grpc-services.adoc[]
-** xref:internal-and-external-communication.adoc[]
 ** Akka Management
 *** Management
 *** Bootstrap
 *** Service Discovery
+
+* How to run Reactive?
+** xref:services.adoc[]
+*** xref:service-discovery.adoc[]
+*** xref:service-clients.adoc[]
+*** xref:grpc-services.adoc[]
 ** Docker
 ** Kubernetes
 ** Helm
--- a/docs-source/docs/modules/concepts/nav.adoc
+++ b/docs-source/docs/modules/concepts/nav.adoc
@@ -1,5 +1,5 @@
-.xref:index.adoc[Concepts]
-* What is Reactive?
+.xref:index.adoc[About Reactive]
+* Knowing Reactive
 ** xref:isolation.adoc[]
 ** xref:autonomy.adoc[]
 ** xref:mobility.adoc[]
@@ -7,7 +7,7 @@
 ** xref:message-driven-event-driven.adoc[Event-Driven]
 ** xref:state-ownership.adoc[]
 
-* How to design Reactive?
+* Designing Reactive
 ** xref:slicing-the-monolith.adoc[]
 ** xref:event-sourcing.adoc[]
 ** xref:cqrs.adoc[]
@@ -21,7 +21,7 @@
 ** xref:ddd.adoc[Brief Summary of DDD concepts]
 ** xref:internal-and-external-communication.adoc[]
 
-* What is in Akka for Reactive?
+* Akka powers Reactive
 ** Actor model
 *** Supervision
 ** Akka Streams
@@ -35,7 +35,7 @@
 *** Bootstrap
 *** Service Discovery
 
-* How to run Reactive?
+* Deploying Reactive
 ** xref:services.adoc[]
 *** xref:service-discovery.adoc[]
 *** xref:service-clients.adoc[]
--- a/docs-source/docs/modules/concepts/nav.adoc
+++ b/docs-source/docs/modules/concepts/nav.adoc
@@ -1,5 +1,5 @@
 .xref:index.adoc[Concepts]
-* Properties of Reactive Microsystems
+* Properties of Reactive
 ** xref:isolation.adoc[]
 ** xref:autonomy.adoc[]
 ** xref:mobility.adoc[]
@@ -7,7 +7,7 @@
 ** xref:message-driven-event-driven.adoc[Event-Driven]
 ** xref:state-ownership.adoc[]
 
-* Patterns and techniques recommended for Reactive Microsystems
+* Patterns and techniques for Reactive
 ** xref:slicing-the-monolith.adoc[]
 ** xref:event-sourcing.adoc[]
 ** xref:cqrs.adoc[]
@@ -20,7 +20,7 @@
 *** Backpressure
 ** xref:ddd.adoc[Brief Summary of DDD concepts]
 
-* Implementation Concepts
+* Tools for Reactive
 ** Actor model
 *** Supervision
 ** Akka Streams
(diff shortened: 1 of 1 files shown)


rstento (approved):
I like these tweaks!

docs-source/docs/modules/concepts/nav.adoc
```diff
@@ -1,13 +1,13 @@
...
+* Properties of Reactive
 ** xref:isolation.adoc[]
 ** xref:autonomy.adoc[]
 ** xref:mobility.adoc[]
 ** Single Responsibility
 ** xref:message-driven-event-driven.adoc[Event-Driven]
 ** xref:state-ownership.adoc[]
 
-* Patterns and techniques recommended for Reactive Microsystems
+* Patterns and techniques for Reactive
```
patriknw, line 10: What is the difference between patterns and techniques here? Would it be enough to use "Patterns for Reactive" or  "Techniques for Reactive"?

docs-source/docs/modules/concepts/nav.adoc
```diff
@@ -1,13 +1,13 @@
 .xref:index.adoc[Concepts]
-* Properties of Reactive Microsystems
+* Properties of Reactive
```
patriknw, line 2: is the removal of Microsystems to make it shorter in the nav?

patriknw: could these be phrased as questions (what is in it for the user) as suggested by @rstento ?

What is Reactive? might work for this one?

How to design Reactive? for next one?

What is in Akka for Reactive? for the third?

I'm blank on the fourth. Have to check what is in there.

docs-source/docs/modules/concepts/nav.adoc
```diff
@@ -1,13 +1,13 @@
...
 ** xref:isolation.adoc[]
 ** xref:autonomy.adoc[]
 ** xref:mobility.adoc[]
 ** Single Responsibility
 ** xref:message-driven-event-driven.adoc[Event-Driven]
 ** xref:state-ownership.adoc[]
```
patriknw, line 8: depending on the navigation theme capabilities we might want to remove this third level in the nav, but that's for later

docs-source/docs/modules/concepts/nav.adoc
```diff
@@ -30,7 +30,7 @@
 ** Akka Persistence
 ** Akka Projections
 
-* Deployment Concerns
+* Deploying Reactive
```
patriknw, line 33: Looking at the current pages in the deploying section many of them are really about deployment in my opinion. More like they would fit in the Tools section. Aside from Akka Management, Docker, Kubernetes.

docs-source/docs/modules/concepts/nav.adoc
ignasi35: I wanted a word that represents the fact that this section lists concepts that are general purpose in software, as in, there's nothing super specific to reactive or microservices but reactive microservices benefit from each of them. 

I'm not too happy calling it _only_ "Patterns for Reactive" (has a strong smell of GoF Patterns). Probably "Techniques for Reactive" or "Useful <something> for Reactive" or "Common in Reactive" or ... 

I'm open to suggestions.

docs-source/docs/modules/concepts/nav.adoc
ignasi35: These three titles sound very appealing. 

The fourth section is a bit of a _miscellanous_ where I put some pages I'm not too sure where else to put yet.

docs-source/docs/modules/concepts/nav.adoc
patriknw: My thought was that "Reactive" was too vague and unspecific. The user is here to learn about Akka Microservices (at least that is the initial scope). "Reactive Systems" is more specific but I kind of like "Reactive Microservices" here.

docs-source/docs/modules/concepts/nav.adoc
ignasi35: Raised #185 to do a pass at Nav later

docs-source/docs/modules/concepts/nav.adoc
ignasi35: Just had a similar thought and moved them around. See the latest commit.

ignasi35: I'm not too attached to the current distribution. I keep reviewing and reordering...

docs-source/docs/modules/concepts/nav.adoc
ignasi35: Good point. Again, I'd revisit this in #185 (once the new theme and more contents have landed).

docs-source/docs/modules/concepts/nav.adoc
```diff
@@ -1,13 +1,13 @@
 .xref:index.adoc[Concepts]
```
patriknw, line 1: by the way, with these 3 or 4 high level groups we could omit the Concepts from the top level and have the others at top level, which is what I think Ruth suggested originally

ignasi35: Hmm, if we remove the `Concepts` umbrella we end up with something like:

Screenshot 2020-09-10 at 19 32 45
Screenshot 2020-09-10 at 19 32 56

I think the `tutorial` and the `how-to` would loose some weight.

ignasi35: Did I understand you correctly ?

patriknw: Correct. Since we don't have too many top level items anyway I think that is good. And might please those that don't like the "Concepts" classification. I don't mind being clear about that it's Concepts but both Ruth and Wade had opinions against it.

The "How to run Reactive?" is still not great. Also misleading to have two "How to" but they are very different in style and content. We should reserve "How to" for the hands-on how to guides.

ignasi35: My concern is that in the view with all items collapsed:

Screenshot 2020-09-10 at 19 32 45

The `tutorial` and the `how to... with Akka` become less relevant.

-----

> The "How to run Reactive?" is still not great. Also misleading to have two "How to" but they are very different in style and content. We should reserve "How to" for the hands-on how to guides.

I know, but I can't think of many other alternatives that are small enough and fit the pattern. Note also the `How to design Reactive?`.

------

Here's yet another possible approach:

Screenshot 2020-09-11 at 09 34 36

It:

1. moves away from the `What/How/Why...?` format to avoid overloading the `how to...` form
2. brings back an umbrella (`About Reactive`) over the four sections to increase the relevance of `tutorial` and `howto`
3. introduces a `<gerund>-ing Reactive` (except for the Akka section, where I break the pattern on purpose)

ignasi35:
I'll mark this DRAFT and maybe we can revisit it later.

rstento:
It is always a good thing to re-visit the titles once most of the content is there. Then, it is not too difficult to rename files and change titles. It is important for SEO that the name of the source file contains the most important words from the title.

docs-source/docs/modules/concepts/nav.adoc
octonato: I like the idea of avoiding a section called "Concepts". "About Reactive" looks better, IMO. 
Also, I like the use of gerund, but I find "Deploying Reactive" a stretch. 

Whatever we put in "Deploying" section is not limited to reactive. Actually, deployment should not be different if the application is reactive or not. 
Deployment is impacted my monitoring, health-checks, if it's clustered service or not, but all of this are valid for any applications, not only reactive applications. So, maybe we will be conflating things here. 

Although I prefer "About Reactive" over "Concepts", I would also be fine with removing the top-level category.

For instance: 
```
 - What is Reactive?
 - Designing Reactive with Akka
 - Deploying Akka applications
 - Tutorial: Microservices
 - How to ... with Akka
```

Note that we can collapse "Designing" with "What is Akka for Reactive" in one section. Then, whenever we cover a topic, we explain how to do it using Akka. That will also help us decide if a topic should be covered or not. If it's something that we can't cover using Akka (because we don't have APIs for or because it doesn't make sense), then we should not cover it in the guide. That said, whenever we bring up a concept, we should be able to bound it to something in Akka. 

About reducing the weight of `tutorial` and `howto`, maybe we could see it differently. 
Both are about hands-on experience. Good if they appear after the theoretical sections "About Reactive" (or flatten structure). 

I don't know if that diminished their value or make then hard to find. Users looking for tutorials/howtos will certainly find it because their eyes are scanning the tree for that. Most probably they will reach the tutorial/howtos pages via a link after an internet search. Moreover, we can give prominence to them in the "Overview" page. 

Alternatively, we can style then if we want to make then more visible. Although, I don't think it's needed.

ignasi35: All the comments above focus on keeping the concepts more or less relevant in the NAV.

There's a completely different option which is to consider the contents we refer to as `concepts` _just_ as supporting material for the `tutorial` and the `how-to`. So, instead of writing a whole bok of concepts and give the impression that users must read it, we use the concepts as a tool to keep the tutorials concise.

With that idea in mind, the NAV simplifies. For example:

```
 - What is Reactive?
 - Designing Reactive with Akka <-- a single item in the NAV (it doesn't even expand)
 > Tutorial: Microservices
 > How to ... with Akka
```

Then, clicking on `Designing Reactive with Akka` would take you to an intro and some links to the most relevant concepts but the weight of the concepts in the NAV is very small.

ignasi35:
Closing in favor of #343

PR #187, opened 2020-09-10
Remove 00-shpopping-cart (use seed instead)

ignasi35:
The `00-shopping-cart-...` codebase is not included and with https://github.com/akka/akka-microservices-seed-scala.g8 now available it looks unnecessary.


patriknw (approved):
Right

Issue #61, opened 2020-08-21
Tutorial source code download


patriknw:
Even though the tutorial will be thorough and include all source code I think we should provide a download of the full source code for each part. I'm thinking that a zip file with different directories for each part would be nice.

ennru:
The packager for this could possibly filter out the tags used for documentation snippets.

ignasi35:
> The packager for this could possibly filter out the tags used for documentation snippets.

Why? We don't do such filtering on any other sample/docs we have. There are some `.g8` templates that include `// # some-tag` comments around the code. Are antora tags more verbose/annoying?

patriknw:
Itt's not too bad, but can be somewhat confusing for users, so nice to filter.

```
// tag::events[]
...
// end::events[]
```

patriknw:
Each step should describe in the beginning that the starting point is the codebase as presented in the previous step, or link to download the zip of previous step.

Before Running in each step we should also have a link to the download in case they failed to follow the instructions (or just not want to).

octonato:
This is done for Scala, but keeping this open since we need for java as well

octonato:
For the record:

I noticed that the tags are not being removed. The sed command we are using is bogus, at least in Mac. 

If it's working on travis, it's fine and we should not spent time on it.

patriknw:
I think it works on Travis

PR #188, opened 2020-09-10
PoC to download the sources

ignasi35:
References #61

To download sources we have some options:

1. (this PR) manually zip, add to git and use antora/asciidoc relative links to [`attachments`](https://docs.antora.org/antora/2.3/asciidoc/link-attachment/)
2. automatically (GitHub action?) build zips and attach the binaries to a GitHub release. HTML uses absolute URLs with the necessary placeholder (tag version, etc...)

PS: To be solved in a separate PR [`filter out the tags`](https://github.com/akka/akka-platform-guide/issues/61#issuecomment-679015361)


patriknw:
Another option is to use TechHub in same way as for akka-samples, ie it creates a zip from a directory. That could also be on a branch if we don’t want to publish each commit and would like to do some scriped pre-prosessing such as removing the doc tags.

docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -118,6 +118,10 @@ You can stop the service with `ctrl-c`.
 
 ifdef::todo[TODO: deploy to cloud and try it]
 
+== Get the source code
+
+link:{attachmentsdir}/01-shopping-cart-service-scala.zip[Download complete project for this step] and run it.
```
patriknw, line 123: I was originally thinking that we would have one single zip with directories for each step. Pros, cons?

ignasi35: Having a single ZIP will simplify things on our end but may confuse users. Using more specific downloads will reduce the chances of mistakes (¿increasing engaging?). Can't think of many other pro/cons.

I thought of the use case where a user only interested in `step 3: persistent entity` could start off using a download that contains _only_ what they need.

So, there would potentially be links to download code at:

1. the beginning of the `step 3: persistent entity` (only has gRPC code)
2. the beginning of each of the projection steps (has gRPC and the complete entity)
3. the end of each of the projection steps (the complete shopping cart and, if applicable, one of the other services)

patriknw: Maybe the simplest thing we can do is to expand the doc deploy script to zip (and cleanup) the sources, and upload them together with the documentation.

ignasi35: > Maybe the simplest thing we can do is to expand the doc deploy script to zip (and cleanup) the sources, and upload them together with the documentation.

SGTM. Makes the site less dependant on external tools (e.g. techhub)

ignasi35: Done in #240

ignasi35:
Closed in favor of

Issue #168, opened 2020-09-09
[Concepts] DDD basic concepts


ignasi35:

patriknw:
I believe this was done

PR #191, opened 2020-09-11
[Concepts] DDD

ignasi35:
References #168

### DIFF
--- a/docs-source/docs/modules/concepts/pages/ddd.adoc
+++ b/docs-source/docs/modules/concepts/pages/ddd.adoc
@@ -26,8 +26,15 @@ Business specialists and other members of the team (both technical and non-techn
 == Stateful Aggregate
 [#aggregate]
 
-An Aggregate is a root entity, the entry point to an object graph of related entities. Using the Aggregate as the unique reference to all the related entities it is possible to guarantee the consistency.
+The Aggregate is a graph of related domain objects that can only be mutated in an atomic operation. The Aggregate is responsible for keeping the invariants of the model. It decides which mutations are allowed according to its business rules and current state.
 
+It models a central concept in your system having a prominet domain object that gives name to it. We call that object the Aggregate Root. 
+
+We can take a shopping cart application as an example. A shopping cart aggregate may contain different items. An item itself is a domain concept, but the shopping cart is the central domain concept in that case. The shopping cart is therefore the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate. 
+
+All operations must pass through the Shopping Cart so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout nor it should be allowed to checkout an empty cart. The shopping cart aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed. 
+
+ 
 == Anti-Corruption Layer
 [#anti_corruption_layer]
 
--- a/docs-source/docs/modules/concepts/pages/ddd.adoc
+++ b/docs-source/docs/modules/concepts/pages/ddd.adoc
@@ -39,7 +39,7 @@ Spliting your business domain model into <<Bounded Context>>'s will provide some
 
 In parallel, you can use xref:event-sourcing.adoc[] to model the <<Stateful Aggregate>>'s on each context.
 
-Both external event-driven modeling and xref:event-sourcing.adoc[] are a perfect match. You can use an <<Anti-Corruption Layer>> and directly publish the Aggregate events for cross-context consumption.
+Both external event-driven modeling and xref:event-sourcing.adoc[] are a perfect match. You can use an <<Anti-Corruption Layer>> and directly publish the Domain events for cross-context consumption.
 
 
 [discrete]
--- a/docs-source/docs/modules/concepts/pages/ddd.adoc
+++ b/docs-source/docs/modules/concepts/pages/ddd.adoc
@@ -19,6 +19,8 @@ Keeping each model within strict boundaries allows different modelling for busin
 
 A model within these bounds may need to be shared to interact with external contexts. Use an <<Anti-Corruption Layer>> to prevent coupling both contexts.
 
+When building Microservices, each service is a Bounded Context encapsulating an autonomous domain model and its external APIs play the role of an Anti-Corruption Layer.
+
 == Ubiquitous Language
 
 Business specialists and other members of the team (both technical and non-technical) should collaborate into creating a single, meaningful naming for the use cases, entities, actions and events modelling the business. Using a single, common language will align all parties and improve communication.
@@ -30,11 +32,11 @@ The Aggregate is a graph of related domain objects that can only be mutated in a
 
 It models a central concept in your system having a prominent domain object that gives name to it. We call that prominent domain object the Aggregate Root. 
 
-We can take a Shopping Cart application as an example. A Shopping Cart may contain different Items (or Products the user buys). An Item itself is a domain concept, but the Shopping Cart is the root domain concept. The Shopping Cart is the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate. 
+We can take the xref:microservices-tutorial:entity.adoc#event_sourced_entity[Shopping Cart] application as an example. A Shopping Cart may contain different Items (or Products the user buys). An Item itself is a domain concept, but the Shopping Cart is the root domain concept. The Shopping Cart is the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate.
 
-All operations must pass through the Shopping Cart so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout nor it should be allowed to checkout an empty cart. The Shopping Cart Aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed. 
+All operations must pass through the Shopping Cart, so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout, nor it should be allowed to checkout an empty cart. The Shopping Cart Aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed.
 
- 
+
 == Anti-Corruption Layer
 [#anti_corruption_layer]
 
--- a/docs-source/docs/modules/concepts/pages/ddd.adoc
+++ b/docs-source/docs/modules/concepts/pages/ddd.adoc
@@ -30,7 +30,7 @@ The Aggregate is a graph of related domain objects that can only be mutated in a
 
 It models a central concept in your system having a prominent domain object that gives name to it. We call that prominent domain object the Aggregate Root. 
 
-We can take a shopping cart application as an example. A shopping cart aggregate may contain different items. An item itself is a domain concept, but the shopping cart is the central domain concept in that case. The shopping cart is therefore the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate. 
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,41 @@
...
 == Anti-Corruption Layer
+[#anti_corruption_layer]
+
+Sharing entities across <<Bounded Context>>'s adds coupling since a change on the entity model affects the code consuming it. An Anti-Corruption Layer is a transformation mapping the external model (controlled by a third party bounded context) into an internal model that's stable to the actual bounded context consuming the model.
+
 == Events-First DDD
+
+ifdef::todo[TODO: this page is WIP]
+
+[discrete]
```
ignasi35, line 39: TIL [`[discrete]`](https://gitter.im/antora/users?at=5f5a8cc2ce5bbc7ffdb84e8f):

> ... it is not included in the table of contents

So I'm using it in order to make `See also` a heading but avoid polluting the TOC.

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,47 @@
...
+
 == Ubiquitous Language
-== Consistency Boundaries
+
+Business specialists and other members of the team (both technical and non-technical) should collaborate into creating a single, meaningful naming for the use cases, entities, actions and events modelling the business. Using a single, common language will align all parties and improve communication.
+
 == Stateful Aggregate
+[#aggregate]
+
+An Aggregate is a root entity, the entry point to an object graph of related entities. Using the Aggregate as the unique reference to all the related entities it is possible to guarantee the consistency.
```
renatocaval, line 29: I think we should re-think how we want to define an aggregate. Strange enough, it's one of the most hard to grab concepts. Each person understand it differently. 

So, I will drop here how I understand it. 

The aggregate is a model that represents a consistency boundary. All operations on that aggregate must be atomic. It's called an aggregate because it can 'aggregate' one or more entities that must be mutated all together.

The term aggregate root means that there is one entity that is prominent and gives the name to the aggregate. A shopping cart is a good example. The root is the `ShoppingCart`, but it can have other entities inside it. 

So the definition in DDD is that you can have one or more entities grouped that participate on the same transaction. However, an entity is also defined as an independent thing that can be referenced directly. So, although they are updated together (transactional boundary), they can be read independently because they can all be referenced by their unique id. Note that there is no distinction between write and read models.

Now, when we bring it to the context of ES and CQRS, the aggregate only exist on the write side and you can only reach the inner entities via the root.

Anyway, I find it's an important concept but hard to delineate. 

We should not say that the "aggregate is the root entity". The aggregate is the whole and we often name if after its root entity.

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,47 @@
...
+
+Sharing entities across <<Bounded Context>>'s adds coupling since a change on the entity model affects the code consuming it. An Anti-Corruption Layer is a transformation mapping the external model (controlled by a third party bounded context) into an internal model that's stable to the actual bounded context consuming the model.
+
 == Events-First DDD
+
+Spliting your business domain model into <<Bounded Context>>'s will provide some spatial decoupling. Then, you can use asynchronous communication between them to achieve some temporal decoupling. Implement that asynchronous communication using xref:message-driven-event-driven.adoc[event-driven] design.
+
+In parallel, you can use xref:event-sourcing.adoc[] to model the <<Stateful Aggregate>>'s on each context.
+
+Both external event-driven modeling and xref:event-sourcing.adoc[] are a perfect match. You can use an <<Anti-Corruption Layer>> and directly publish the Aggregate events for cross-context consumption.
```
renatocaval, line 42: maybe better to call it "Domain events"

```suggestion
Both external event-driven modeling and xref:event-sourcing.adoc[] are a perfect match. You can use an <<Anti-Corruption Layer>> and directly publish the Domain events for cross-context consumption.
```

renatocaval (commented):
Looking really good. 

I did a first pass. If will be good to have some more iterations.

ignasi35:
I'm all for improving it and the more hands reviewing/extending, the better. Still, this is low value on the guide. DDD is tangentially necessary for the main purpose of the guide so maybe we can keep this content as is and mark future improvements as low-priority for the time being?

I found myself cross-referencing Event Sourcing, Events vs Messages, and some other concepts from multiple pages, but that was barely the case with the concepts in this DDD intro. :shrug:

docs-source/docs/modules/concepts/pages/ddd.adoc
ignasi35: @renatocaval feel free to take over the PR with the improved writeup you suggest in this comment. :-)

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,54 @@
...
+
+It models a central concept in your system having a prominet domain object that gives name to it. We call that object the Aggregate Root. 
+
+We can take a shopping cart application as an example. A shopping cart aggregate may contain different items. An item itself is a domain concept, but the shopping cart is the central domain concept in that case. The shopping cart is therefore the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate. 
+
+All operations must pass through the Shopping Cart so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout nor it should be allowed to checkout an empty cart. The shopping cart aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed.
```
octonato, line 35: @ignasi35, I re-worked this paragraph as requested. Let me know what you think.

ignasi35 (commented):
The Shopping Cart examples are useful but, compared to the other sections on the page, too detailed. I'm not sure we should go into that level of detail.

octonato:
> The Shopping Cart examples are useful but, compared to the other sections on the page, too detailed. I'm not sure we should go into that level of detail.

I also thought that it could be too detailed. On the other hand, it makes sense to have a concrete example at hand to explain what an aggregate it. Otherwise it's too abstract. 

If we agree to keep as is, we can cross-link.

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,54 @@
...
+
+It models a central concept in your system having a prominent domain object that gives name to it. We call that prominent domain object the Aggregate Root. 
+
+We can take a Shopping Cart application as an example. A Shopping Cart may contain different Items (or Products the user buys). An Item itself is a domain concept, but the Shopping Cart is the root domain concept. The Shopping Cart is the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate. 
+
+All operations must pass through the Shopping Cart so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout nor it should be allowed to checkout an empty cart. The shopping cart aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed.
```
octonato, line 35:
```suggestion
All operations must pass through the Shopping Cart so it can check its validity. You may add or remove an item, or change its quantity, you can checkout the cart. All those operations are atomic and sequential. It should not be allowed to change its content at the same time the cart is being checkout nor it should be allowed to checkout an empty cart. The Shopping Cart Aggregate will act as the guardian of the model integrity ensuring that only valid mutations are allowed. 
```

ignasi35:
> > The Shopping Cart examples are useful but, compared to the other sections on the page, too detailed. I'm not sure we should go into that level of detail.
> 
> I also thought that it could be too detailed. On the other hand, it makes sense to have a concrete example at hand to explain what an aggregate it. Otherwise it's too abstract.
> 
> If we agree to keep as is, we can cross-link.

You're right, even if it makes this section on the page more detailed than the rest it's also an indicator of how much more important the `Aggregate` is compared to other concepts introduced in the `ddd.adoc` page. 

Let's merge as is and we can revisit in the future.

ignasi35:
Meh, I submitted the PR originally so I can't `Approve` it.

octonato:
I will cross-link and then we merge.

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,54 @@
...
+____
 
 == Bounded Context
 [#bounded_context]
+
+As the modelling covers a more extensive area of your business, split the models into very well-delimited contexts in which each model applies. Each of these contexts will have autonomy to evolve the models it owns.
+
+Keeping each model within strict boundaries allows different modelling for business entities that look similar but have slightly different meaning on each of the contexts.
+
+A model within these bounds may need to be shared to interact with external contexts. Use an <<Anti-Corruption Layer>> to prevent coupling both contexts.
```
patriknw, line 20: should we mention how this relates to Microservices? Each Microservice has it's own bounded context.

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,54 @@
...
+Business specialists and other members of the team (both technical and non-technical) should collaborate into creating a single, meaningful naming for the use cases, entities, actions and events modelling the business. Using a single, common language will align all parties and improve communication.
+
 == Stateful Aggregate
+[#aggregate]
+
+The Aggregate is a graph of related domain objects that can only be mutated in an atomic operation. The Aggregate is responsible for keeping the invariants of the model. It decides which mutations are allowed according to its business rules and current state.
+
+It models a central concept in your system having a prominent domain object that gives name to it. We call that prominent domain object the Aggregate Root. 
+
+We can take a Shopping Cart application as an example. A Shopping Cart may contain different Items (or Products the user buys). An Item itself is a domain concept, but the Shopping Cart is the root domain concept. The Shopping Cart is the Aggregate Root and gives name to the Aggregate. We can simply talk about the Shopping Cart Aggregate.
```
patriknw, line 33: good to connect to the example

docs-source/docs/modules/concepts/pages/ddd.adoc
```diff
@@ -5,14 +5,54 @@
...
+
+Spliting your business domain model into <<Bounded Context>>'s will provide some spatial decoupling. Then, you can use asynchronous communication between them to achieve some temporal decoupling. Implement that asynchronous communication using xref:message-driven-event-driven.adoc[event-driven] design.
+
+In parallel, you can use xref:event-sourcing.adoc[] to model the <<Stateful Aggregate>>'s on each context.
+
+Both external event-driven modeling and xref:event-sourcing.adoc[] are a perfect match. You can use an <<Anti-Corruption Layer>> and directly publish the Domain events for cross-context consumption.
+
+
+[discrete]
+== See also
```
patriknw, line 53: would it make sense to use the same as in the tutorial, which is "Learn more" ?

patriknw (approved):
LGTM, two optional suggestions

docs-source/docs/modules/concepts/pages/ddd.adoc
ignasi35: Now that you mention it, I think I've been a bit inconsistent. We can do a complete pass on a separate PR.

ignasi35: raised https://github.com/akka/akka-platform-guide/issues/235

docs-source/docs/modules/concepts/pages/ddd.adoc
ignasi35: +1 @octonato do you mind adding this bit when you add the x-ref from the aggregate to the tutorial? Thanks! :-)

octonato: I will do it.

octonato:
@patriknw / @ignasi35, I did the cross-link and related bounded context with microservices. 

Please, another review round.

patriknw:
new changes lgtm

ignasi35 (commented):
LGTM

Issue #185, opened 2020-09-10
Review NAV depth and contents


ignasi35:
See discussion at https://github.com/akka/akka-platform-guide/pull/177#discussion_r486313550:

> I'm a bit undecided about what should the nav look like:
> 
> a) too detailed becomes noisy and messy: detailed concepts mixed with tutorial steps. As content grows, this alternative make snavigation too difficult aswell.
> b) too scarce makes navigation more difficult

patriknw:
I think this is handled by https://github.com/akka/akka-platform-guide/pull/343 for now. We might have to revisit when adding more How to, but that will probably be handled by not showing all those in nave but in an index page instead.

PR #192, opened 2020-09-11
Scaffolds a page for each concept

ignasi35:
References #59

Related to #185 and #179 

In #184, we've been ping-pong'ing names for the sections under the `Concepts` module. In this PR, while creating an empty page for each of the concepts I did a bit of cleanup and reorganized the concepts into only 3 groups:

* Properties of Reactive Microsystems: this hasn't changed.
* Implementation Concepts: all pages introducing implementation details using Akka.
* Patterns and techniques...: all the rest of the concepts. I moved this section to the bottom and also merged it with what I previously called `Deployment concepts`. These changes increase the importance of the Akka section (now second) and simplify the task on #184 

Note: it is not the purpose of this RP to review the names of each section. I only wanted to create empty pages for the missing concepts.


docs-source/docs/modules/concepts/pages/memory-image-pattern.adoc
```diff
@@ -0,0 +1,12 @@
+= Memory Image Pattern
```
patriknw, line 1: not sure if this is needed, isn't that covered "enough" in Event Sourcing?

patriknw (approved):
LGTM

docs-source/docs/modules/concepts/pages/memory-image-pattern.adoc
ignasi35: Probably, I don't know. For the time being, I'm using these empty pages as mnemonics. 

In the past I found myself writing a dedicated page for concepts that turned out to be so small I ended up merging them with existing pages. My concern, if we move Memory Image to Event Sourcing, is that the Event Sourcing page will be a bit too big: "Event Sourcing it _the way_ but you have to learn all this...".

octonato: What is "Memory Image Pattern"? Are we referring to this: https://martinfowler.com/bliki/MemoryImage.html

I wouldn't create a section for that and leave it implicit in the as something we can achieve with `EventSourcedBehavior`. A small mention there, eventually with a link to Martin Fowler article.

octonato (approved):

docs-source/docs/modules/concepts/pages/memory-image-pattern.adoc
ignasi35: A section about memory image is just an excuse to link to MFowler's article *and* explain how Akka Persistence and Akka Sharding are combined to implement it (briefly introducing the responsibilties of each of the Akka modules).

As in: "if you only used Akka Persistence, then blabla" and "sharding is unrelated to keeping data in mem".

Again, not sure where this content will belong to, but just a reminder that we can add it.

PR #200, opened 2020-09-16
bot to generate traffic

patriknw:
Since we can't try the gRPC ingress I created this bot to generate some traffic.
Not to be merged. Just sharing in case someone else would like to use it


raboof:
> Since we can't try the gRPC ingress

For future reference: you can use `kubectl port-forward` to directly talk to the pod, see for example https://cloudflow.io/docs/current/get-started/exercise-example.html for an example of how this is used in cloudflow.

patriknw:
> For future reference: you can use kubectl port-forward to directly talk to the pod

Thanks

PR #201, opened 2020-09-16
use the same CassandraSession for journal and projections

patriknw:
* more efficient and easier to configure since we don't aim for using more than one db


ennru (approved):
LGTM.

johanandren (approved):

Issue #28, opened 2020-08-19
Document evolving proto schemas


ennru:
The documentation should explain the basics of how APIs can be evolved when using Protobuf.

ennru:
This should include how proto files can be distributed to clients.

ennru:
Maybe some directions to enable idempotency.

patriknw:
I assume this ticket also covers "How to evolve gRPC APIs with Akka gRPC". We had an empty page for that. I remove that page and it can be added if/when we do this ticket.

PR #205, opened 2020-09-16
correct proto index

patriknw:
reported by @mckeeh3

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/01-shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto
+++ b/docs-source/docs/modules/microservices-tutorial/examples/01-shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto
@@ -18,7 +18,7 @@ message AddItemRequest { // <2>
 }
 
 message Cart { // <3>
-    repeated Item items = 2;
+    repeated Item items = 1;
 }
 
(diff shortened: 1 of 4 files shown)


ignasi35:
Should [the final file](
https://github.com/akka/akka-platform-guide/blob/main/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto#L44-L47) switch the field order, then?

patriknw:
> Should the final file switch the field order, then?

I don't know if it's only me but I tend to put repeated fields at the end.

I don't think it matters that they differ.

ignasi35:
But with this PR we'd be changing the tag of the `Items` field.
```
// from the final codebase
message Cart {
    bool checkedOut = 1;
    repeated Item items = 2;
}
```
```
// from steps 01 and 02
message Cart {
    repeated Item items = 1;
}
```

ignasi35:
My "protobuf evolutions fu" is veeeery rusted.

ignasi35:
I mean, it's a small thing but users following the tutorial may have `Cart` snapshots the final codebase would not be able to read(?). 
Also, implementing this with evolution of protos in mind will simplify the `How to evolve proto schemas` page (in #28).

ignasi35 (approved):

PR #217, opened 2020-09-29
Add create keyspace to Cassandra script

octonato:
Not clear if it was intentional to leave the `create keyspace` out, so here it is. 

I'm assuming development / test and just using the same settings suggested in https://doc.akka.io/docs/akka-persistence-cassandra/current/journal.html#schema


patriknw (approved):
LGTM, that must have been an overisight

Issue #206, opened 2020-09-16
Consider splitting up the code in Part 4: Complete Event Sourced entity


patriknw:
The intention is that it's more like an exercise for the reader, and then showing the full thing. Maybe that doesn't work. Shall we present it in smaller pieces?

original comment from @mckeeh3

patriknw:
I'd suggest that we show only `Checkout` command/event first, followed by `Get` command. Then the steps/code should be small enough. We can refer to the source download for the optional `RemoveItem` and `AdjustItemQuantity`.

PR #221, opened 2020-10-01
simplify Complete Event Sourced entity, #206

patriknw:
* by only showing the Checkout and Get commands
* and show the solution piece by piece instead of a big wall of code
* for the optional commands we can refer to the source download

References #206

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/ShoppingCartSpec.scala
+++ b/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/ShoppingCartSpec.scala
@@ -70,6 +70,7 @@ class ShoppingCartSpec
       result2.event should ===(ShoppingCart.ItemQuantityAdjusted(cartId, "foo", 43, 42))
     }
 
+    // tag::checkout[]
     "checkout" in {
       val result1 =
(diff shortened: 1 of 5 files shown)


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/ShoppingCartSpec.scala
```diff
@@ -82,6 +83,18 @@ class ShoppingCartSpec
         eventSourcedTestKit.runCommand[StatusReply[ShoppingCart.Summary]](ShoppingCart.AddItem("bar", 13, _))
       result3.reply.isError should ===(true)
     }
+    // end::checkout[]
```
ignasi35, line 86: We can later extend the tests around the `Checkout` command. I thin there's three different situations which should be tested separately:

1. trying to checkout an empty cart should fail
2. trying to checkout a cart with items produces the CheckedOutEvent and a the expected summary
3. trying to checkout a checked-out cart should fail

ignasi35: If we introduce these three business rules in the writeup as a list, it'll be easy to refer to each business rule from each test and production code (especially when splitting the command handler into two functions for each state of the cart).

ignasi35 (approved):
LGTM with some ideas for improvement in a separate PR.

octonato (approved):

Issue #214, opened 2020-09-29
Wrong ActorSystem name in Part 3: First gRPC service


patriknw:
`ActorSystem[Nothing](Main(), "Cart")` should be `ActorSystem[Nothing](Main(), "ShoppingCartService")`

ignasi35:
IIUC, this change is for cluster bootstrap. How about introducing:

```
akka.management {
  cluster.bootstrap {
    contact-point-discovery {
      ...
      service-name = "shoppingcartservice"
    }
  }
}
```

in the settings and decouple the deployment from the actor system name.

patriknw:
We can do that to connect the dots more explicitly. We should still update the ActorSystem name, since I don't think it's matching what the seed generated.

PR #222, opened 2020-10-01
correction of ActorSystem name

patriknw:
* and explicit config of service-name

References #214


docs-source/docs/modules/microservices-tutorial/examples/shopping-analytics-service-scala/src/main/resources/local-shared.conf
```diff
@@ -2,6 +2,7 @@ akka.remote.artery.canonical.hostname = "127.0.0.1"
 akka.management.http.hostname = "127.0.0.1"
 
 akka.management.cluster.bootstrap.contact-point-discovery {
+  service-name = "shoppinganalyticsservice"
```
patriknw, line 5: if we agree that we should have this explicit config instead of the ActorSystem name convention I will update the seed template

ignasi35: +1 from me.

ignasi35 (approved):

patriknw:
I'll add to the seed

patriknw:
https://github.com/akka/akka-microservices-seed-scala.g8/pull/10

PR #227, opened 2020-10-01
[Style] Specify file name

ignasi35:
It is very helpful to always indicate what file a snippet belongs to.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
@@ -25,6 +25,7 @@ The motivation for using gRPC as inter-service protocol is explained in the {akk
 
 First you should define the interface of the service in a protobuf service descriptor. It should be located in the `src/main/protobuf/ShoppingCartService.proto`.
 
+.shopping-cart-service/src/main/protobuf/ShoppingCartService.proto
 [source,protobuf]
 ----
 include::example$01-shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto[]
@@ -46,6 +47,7 @@ sbt compile
 
 Let's implement the `ShoppingCartService` trait. Create a `ShoppingCartServiceImpl` class that extends the generated `ShoppingCartService` trait. Implement the `addItem` method from the trait by logging the call and return a successful `Future`.
 
+.shopping-cart-service/src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala
 [source,scala]
 ----
 include::example$01-shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala[]
@@ -57,6 +59,7 @@ include::example$01-shopping-cart-service-scala/src/main/scala/shopping/cart/Sho
 
 The gRPC service implementation is running by an Akka HTTP server. Add the following server initialization code in a `ShoppingCartServer` object.
 
+.shopping-cart-service/src/main/scala/shopping/cart/ShoppingCartServer.scala
 [source,scala]
 ----
 include::example$01-shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServer.scala[]
@@ -70,6 +73,7 @@ include::example$01-shopping-cart-service-scala/src/main/scala/shopping/cart/Sho
 
 To run the service we need a class with a `main` method. Edit the `Main` class that is included from the template project. It should initialize the `ActorSystem` and the `ShoppingCartServer` like this:
 
+.shopping-cart-service/src/main/scala/shopping/cart/Main.scala
 [source,scala]
 ----
 include::example$01-shopping-cart-service-scala/src/main/scala/shopping/cart/Main.scala[]


docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -25,6 +25,7 @@ The motivation for using gRPC as inter-service protocol is explained in the {akk
 
 First you should define the interface of the service in a protobuf service descriptor. It should be located in the `src/main/protobuf/ShoppingCartService.proto`.
 
+.src/main/protobuf/ShoppingCartService.proto
```
ignasi35, line 28:
```suggestion
.shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto
```

docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -25,6 +25,7 @@ The motivation for using gRPC as inter-service protocol is explained in the {akk
 
 First you should define the interface of the service in a protobuf service descriptor. It should be located in the `src/main/protobuf/ShoppingCartService.proto`.
 
+.shopping-cart-service-scala/src/main/protobuf/ShoppingCartService.proto
```
ignasi35, line 28: In this case, I didn't remove the sentence immediately before the snippet but we could even remove it.

docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -46,6 +47,7 @@ sbt compile
 
 Let's implement the `ShoppingCartService` trait. Create a `ShoppingCartServiceImpl` class that extends the generated `ShoppingCartService` trait. Implement the `addItem` method from the trait by logging the call and return a successful `Future`.
 
+.shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala
```
ignasi35, line 50: We take for granted users are familiar with the `src/main/scala/com/package` folder structure. Adding this metadata will prevent any confusion.

I just had to do a quick scan over the file copy/pasting the relevant bit from the `include::` macro below. Note that I stripped the `01-` prefix from the root folder.

ignasi35:
I only did one page of the tutorial but I'd merge this PR already. If we crowd-source this change or apply the change as we edit other pages it should be enough.

octonato (commented):
I don't know if it's a good idea. 

We are showing the path to file in the full project, but if you are following the guide using the seed template, the name of the project will be `shopping-cart-service`. 

> We take for granted users are familiar with the src/main/scala/com/package folder structure. Adding this metadata will prevent any confusion.

I think we can take that for granted and if not, people can learn new things. We can expect a minimal of autonomy from a developer.

docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
octonato: If we go for this option, I think we should not include `shopping-cart-service-scala`. Starting on `src` is good.

ignasi35:
> We are showing the path to file in the full project, but if you are following the guide using the seed template, the name of the project will be `shopping-cart-service`.

Good catch. Fixed.

ignasi35:
> I don't know if it's a good idea.

Fair enough. Let's see what others think. :-)

patriknw:
I can see the value of this.
In some places we might want to adjust the text if we have this information anyway. For example "It should be located in the src/main/protobuf/ShoppingCartService.proto."

However, I wonder how this will look when we have Scala/Java tabs?

ignasi35:
> However, I wonder how this will look when we have Scala/Java tabs?

Ah, good point. I don't know enough asciidoc or details about the solution Rich is putting together.

Let's put this on-hold.

octonato:
We can add the pass as a comment inside the snippet at the first line. 

Also, it looks bit weird now. Looks like we forgot something in between. 

If we leave it outside, we can label it with something. Could be:

**Location:** `src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala`

ignasi35:
> If we leave it outside, we can label it with something. Could be:
> 
> **Location:** `src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala`

Yeha, the good thing of using the syntax:

```
.path/filename.txt
[source, scala]
----
include...
----
```
Is that it's asciidoc syntax we can then theme however we like.

ignasi35:
Screenshot 2020-10-08 at 11 27 25

rstento:
> I don't know if it's a good idea.
> 
> We are showing the path to file in the full project, but if you are following the guide using the seed template, the name of the project will be `shopping-cart-service`.
> 
> > We take for granted users are familiar with the src/main/scala/com/package folder structure. Adding this metadata will prevent any confusion.
> 
> I think we can take that for granted and if not, people can learn new things. We can expect a minimal of autonomy from a developer.

True, @octonato, but in a tutorial we want to do everything we can to prevent them from making silly mistakes. I believe it is a good thing to show the filename with path. If they don't need to see it, they can ignore it.

rstento:
> > If we leave it outside, we can label it with something. Could be:
> > **Location:** `src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala`
> 
> Yeha, the good thing of using the syntax:
> 
> ```
> .path/filename.txt
> [source, scala]
> ----
> include...
> ----
> ```
> 
> Is that it's asciidoc syntax we can then theme however we like.

Yes, I think that the default styling of the title for a code block is a bit too big and too bold, and I don't like the italics. We can ask @rasummer to style it for us. I will try a few things and let you know.

ignasi35:
Once discussed and agreed, I'm closing this (obsolete PR) in favor of Tracking it as a new issue in #363

Issue #233, opened 2020-10-02
Include docker-compose instructions in dev-env.adoc


octonato:
Shall we also add a new section in `dev-env.adoc` about docker-compose and this script?

_Originally posted by @patriknw in https://github.com/akka/akka-platform-guide/pull/220#discussion_r498051023_

PR #238, opened 2020-10-02
Add docker instructions to dev-env.adoc

octonato:
References #233

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
@@ -67,9 +67,9 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 == Docker and Docker Compose
 [#docker]
 
-This tutorial makes used of Cassandra and Kafka. You may use local installations during development, but we recommend run both servers using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to bootstrap Cassandra, Kafka and Zookeeper (the latter required by the Kafka server).
+This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
 
-After installing `docker` and `docker-compose`, you can run the following command at the root of the project.
+After installing `docker` and `docker-compose`, you can run the following command at the root of the project:
 
 
 // tag::docker[]
--- a/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
@@ -67,7 +67,7 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 == Docker and Docker Compose
 [#docker]
 
-This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
+This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker] https://docs.docker.com[{tab-icon}, window="tab"]. The template project includes a https://docs.docker.com/compose/[Docker Compose]  https://docs.docker.com/compose/[{tab-icon}, window="tab"] file on its root you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
 
 After installing `docker` and `docker-compose`, you can run the following command at the root of the project:
 
--- a/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 3. open the `shopping-cart-service/build.sbt` from __File > Open__.
 4. select __Open as project__
 
-== grpcurl
+== Docker and Docker Compose
+[#docker]
 
-https://github.com/fullstorydev/grpcurl[grpcurl] is useful for making gRPC calls from the command line. It is used in the tutorial and it's recommended that you https://github.com/fullstorydev/grpcurl#installation[install] it.
+This tutorial makes used of Cassandra and Kafka. You may use local installations during development, but we recommend run both servers using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to bootstrap Cassandra, Kafka and Zookeeper (the latter required by the Kafka server).
 
+After installing `docker` and `docker-compose`, you can run the following command at the root of the project.
+
+
+// tag::docker[]
+----
+docker-compose up -d
+----
+
+Create the Cassandra keyspace and tables with:
+
+----
+docker exec -i shopping-cart-service_cassandra_1 cqlsh -t < ddl-scripts/create_tables.cql
+----
+
+This CQL script is located inside the `ddl-scripts` at the root of the project. It will create the keyspace and all tables needed for Akka Persistence as well as the offset store table for Akka Projection.
+
+[CAUTION]
+====
+* The keyspace as created by the script is probably not what you need in a production environment. This is good enough for local development though.
+====
+
+// tag::dockerImportant[]
+[IMPORTANT]
+====
+* When loading the CQL script, make sure to use the same name as your running Cassandra container name. The container name is not fixed and depends on the parent folder of the docker-compose file. The above example assumes the project was created using the seed template and named `shopping-cart-service`.
+
+* If you get a connection error with the message *_Unable to connect to any servers_*, it means the Cassandra container is still starting. Wait a few seconds and re-try the command.
+====
+// end::dockerImportant[]
+// end::docker[]
 
 == Config for local development
 
@@ -149,6 +180,12 @@ The example projects are using the following ports.
 |8303
 
 |===
+
+
+== grpcurl
+
+https://github.com/fullstorydev/grpcurl[grpcurl] is useful for making gRPC calls from the command line. It is used in the tutorial and it's recommended that you https://github.com/fullstorydev/grpcurl#installation[install] it.
+
 
 == Learn more
 
--- a/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
@@ -181,27 +181,7 @@ include::example$02-shopping-cart-service-scala/src/main/resources/application.c
 
 First you need to start Cassandra with:
 
-----
-docker-compose up -d
-----
-
-Create the Cassandra keyspace and tables with: 
-
-----
-docker exec -i shopping-cart-service_cassandra_1 cqlsh -t < ddl-scripts/create_tables.cql
-----
-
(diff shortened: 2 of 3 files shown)


docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -150,6 +181,12 @@ The example projects are using the following ports.
 
 |===
 
+
+== grpcurl
```
octonato, line 185: I changed the order here to follow the order during development.

* start docker
* run locally with sbt
* do some calls with grpcurl

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 3. open the `shopping-cart-service/build.sbt` from __File > Open__.
 4. select __Open as project__
 
-== grpcurl
```
octonato, line 67: moved down, to after starting the server

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
```diff
@@ -221,7 +201,7 @@ Try the validation logic by trying to add the same item again, which should resu
 grpcurl -d '{"cartId":"cart1", "itemId":"socks", "quantity":5}' -plaintext 127.0.0.1:8101 shoppingcart.ShoppingCartService.AddItem
 ----
 
-To verify that the events are actually saved and the state can be recovered from the events you can stop the service with `ctrl-c` and then start it again.
+To verify that the events are actually saved, and the state can be recovered from the events you can stop the service with `ctrl-c` and then start it again.
```
octonato, line 204: This was a grammar recommendation from JetBrains ;-)

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
```diff
@@ -181,27 +181,7 @@ include::example$02-shopping-cart-service-scala/src/main/resources/application.c
...
-[NOTE]
-====
-* The keyspace as created by the script is probably not what you need in a production environment. This is good enough for local development though. 
-
-* When loading the `create_tables.cql` file, make sure to use the same name as your running Cassandra container name. The container name is not fixed and depends on the parent folder of the docker-compose file. The above example assumes the project was created using the seed template and named `shopping-cart-service`.
-
-* If you get a connection error with the message *_Unable to connect to any servers_*, it means the Cassandra container is still starting. Wait a few seconds and re-try the command.
-====
-
+include::dev-env.adoc[tag=docker]
```
octonato, line 184: @patriknw, I didn't add a link to dev-env for troubleshooting, but simply included the whole section. Users don't need to go back and forth.

docs-source/docs/modules/microservices-tutorial/pages/projection-query.adoc
```diff
@@ -154,12 +154,8 @@ Load the file into Cassandra:
 docker exec -i shopping-cart-service_cassandra_1 cqlsh -t < ddl-scripts/create_user_tables.cql
 ----
 
-[NOTE]
-====
-* When loading the `create_user_tables.cql` file, make sure to use the same name as your running Cassandra container name. The container name is not fixed and depends on the parent folder of the docker-compose file. The above example assumes the project was created using the seed template and named `shopping-cart-service`.
 
-* If you get a connection error with the message *_Unable to connect to any servers_*, it means the Cassandra container is still starting. Wait a few seconds and re-try the command.
-====
+include::dev-env.adoc[tag=dockerImportant]
```
octonato, line 158: Here we include only the troubleshooting part, not need to warn about keyspace settings.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
...
+
+This CQL script is located inside the `ddl-scripts` at the root of the project. It will create the keyspace and all tables needed for Akka Persistence as well as the offset store table for Akka Projection.
+
+[CAUTION]
+====
+* The keyspace as created by the script is probably not what you need in a production environment. This is good enough for local development though.
+====
 
+// tag::dockerImportant[]
+[IMPORTANT]
```
octonato, line 94: Two kinds of admonitions as defined here https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#paragraphs

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 3. open the `shopping-cart-service/build.sbt` from __File > Open__.
 4. select __Open as project__
 
-== grpcurl
+== Docker and Docker Compose
+[#docker]
 
-https://github.com/fullstorydev/grpcurl[grpcurl] is useful for making gRPC calls from the command line. It is used in the tutorial and it's recommended that you https://github.com/fullstorydev/grpcurl#installation[install] it.
+This tutorial makes used of Cassandra and Kafka. You may use local installations during development, but we recommend run both servers using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to bootstrap Cassandra, Kafka and Zookeeper (the latter required by the Kafka server).
```
ignasi35, line 70:
```suggestion
This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
```

ignasi35: Ths links to docker and docker-compose are valid asciidoc syntax but don't adhere to the style required for external links:

    https://some-domain.org/foo[Link text] https://some-domain.org/foo[{tab-icon}, window="tab"]

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
...
+
+After installing `docker` and `docker-compose`, you can run the following command at the root of the project.
```
ignasi35, line 72: It'd be good to link to the installation instructions of both docker and docker-compose here.

ignasi35, line 72: > [...] you can run the following command at the root of the project

I can run that command, but why? ;-)

PS. also, when introducing a snippet, especially if using the word "following" prefer ending the sentence with a colon `:`.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
...
+
+
+// tag::docker[]
+----
+docker-compose up -d
+----
+
+Create the Cassandra keyspace and tables with:
```
ignasi35, line 80: Users that preferred _not_ using docker compose, may not see this as a required step.

IMHO, creating the tables in the DB should not be under the `== Docker and Docker Compose` section but a  dedicated one, instead.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
ignasi35: I also prefer the syntax:

```
[FOO]
=====
some text
====
```

for admonitions.

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
ignasi35: I think it adds too much noise. I'd rather right-click, do what I need to do, and close the new tab.

docs-source/docs/modules/microservices-tutorial/pages/projection-query.adoc
ignasi35: This `include`, OTOH, looks appropriate since it'll help prevent issues.

ignasi35 (changes requested):

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
octonato: What do you mean why? Do you think we should explain why docker is a good idea?

I would rather not explain everything. If we say that we recommend something and provide the links, and users have no idea what is that, they can ignore the recommendation or research, no?

Otherwise we will add to much information in our guide. Most people are not already familiar with docker and will understand why we recommend it.

octonato, line 72:
```suggestion
After installing `docker` and `docker-compose`, you can run the following command at the root of the project:
```

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
octonato: This is still a required step for them. We want to add this here because we want to show how to load it in docker. 

If a user prefers to run Cassandra and Kafka locally, they are on their own. I would assume that they are comfortable enough to see `ddl-scripts/create_tables.cql` and understand that they should pass it to cqlsh.

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
octonato: I don't know. The goal was exactly to avoid context switching. 

Later on, I also add this again when creating the projection table, but only the troubleshooting part. The instructions are slightly different and it doesn't make sense to put a link. So, I opt to include the full instruction each time, so we have homogeneity.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -64,10 +64,41 @@ You can use any code editor you prefer for this tutorial. We recommend https://w
 3. open the `shopping-cart-service/build.sbt` from __File > Open__.
 4. select __Open as project__
 
-== grpcurl
+== Docker and Docker Compose
+[#docker]
 
-https://github.com/fullstorydev/grpcurl[grpcurl] is useful for making gRPC calls from the command line. It is used in the tutorial and it's recommended that you https://github.com/fullstorydev/grpcurl#installation[install] it.
+This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker]. The template project includes a https://docs.docker.com/compose/[Docker Compose] file on its root you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
```
octonato, line 70: note to self: change the links to open on another tag

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
ignasi35: Hmmm, I see. Let me have a second look. I'll actually run the site to get a better feel.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
ignasi35: I meant that the following sentence may be a bit more understandable:

```
After installing `docker` and `docker-compose`, use the following command at the root of the project to start Cassandra and Kafka:  
```

What I meant with `but why?` is that we either assume they know nothing and explain everything in detail or we assume they are familiar with docker/docker-compose and don't even bother detailing the command.

ignasi35 (approved):

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
ignasi35: @octonato I'll do that and push to your branch.

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
patriknw: I think it's ok to repeat. It's not in each and every Run section.

patriknw (approved):
LGTM

PR #225, opened 2020-10-01
Split up the Scala example in parts, #199

patriknw:
No point in reviewing full PR. Look at the "modified" commit, and the `copy-identical-files.sh` script.

References #199

PR #240, opened 2020-10-05
Adds a scripts to produce the downloadable zip files

ignasi35:
Supersedes #188 
References #61
Blocked by #225

### DIFF
--- a/scripts/prepare-dowloads.sh
+++ b/scripts/prepare-dowloads.sh
@@ -3,7 +3,7 @@
 declare -r tutorial_sources="${PWD}/docs-source/docs/modules/microservices-tutorial/examples"
 declare -r tutorial_attachments="${PWD}/docs-source/docs/modules/microservices-tutorial/assets/attachments"
 
-declare -r temporal_folder="${PWD}target/zips"
+declare -r temporal_folder="${PWD}/target/zips"
 
 mkdir -p ${tutorial_attachments}
 
--- a/scripts/prepare-dowloads.sh
+++ b/scripts/prepare-dowloads.sh
@@ -49,6 +49,12 @@ function zipAndAttach() {
 removeTags
 
 
+## empty codebase (this is a backup option for users 
+## that can't use the .g8 template)
+prepareTemporalFolder
+fetchProject 00-shopping-cart-service-scala shopping-cart-service
+zipAndAttach step-2.zip
+
 ## gRPC service
 prepareTemporalFolder
 fetchProject 01-shopping-cart-service-scala shopping-cart-service
--- a/scripts/prepare-dowloads.sh
+++ b/scripts/prepare-dowloads.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+declare -r tutorial_sources="docs-source/docs/modules/microservices-tutorial/examples"
+declare -r tutorial_attachments="docs-source/docs/modules/microservices-tutorial/assets/attachments"
+
+function prepareAttachment() {
+   source_name=$1
+   zip ${tutorial_attachments}/${source_name}.zip ${tutorial_sources}/${source_name}
+}
+
+mkdir -p ${tutorial_attachments}
+
+prepareAttachment 00-shopping-cart-service-scala
+prepareAttachment 01-shopping-cart-service-scala
+prepareAttachment 02-shopping-cart-service-scala
+prepareAttachment 03-shopping-cart-service-scala
+prepareAttachment 04-shopping-cart-service-scala
+prepareAttachment 05-shopping-cart-service-scala
+prepareAttachment shopping-cart-service-scala
+prepareAttachment shopping-analytics-service-scala
+prepareAttachment shopping-order-service-scala
--- a/scripts/prepare-dowloads.sh
+++ b/scripts/prepare-dowloads.sh
@@ -7,17 +7,25 @@ declare -r temporal_folder="${PWD}/target/zips"
 
 mkdir -p ${tutorial_attachments}
 
+## Cleanup the temporal folder from previous executions
 function prepareTemporalFolder() {
    rm -rf ${temporal_folder}
    mkdir -p ${temporal_folder}
 }
 
+## Copy a folder with some code into the temporal folder. The 
+## copied folder will be renamed to the folder name we want the 
+## user to see when unzipping the file.
+##   source_name -> subfolder in `microservices-tutorial/examples`
+##   target_name ->  folder name the user should see (must not use a numeric prefix of a laguage suffix)
 function fetchProject() {
    source_name=$1
    target_name=$2
    cp -a ${tutorial_sources}/${source_name} ${temporal_folder}/${target_name}
 }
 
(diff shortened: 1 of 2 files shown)


scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+declare -r tutorial_sources="docs-source/docs/modules/microservices-tutorial/examples"
+declare -r tutorial_attachments="docs-source/docs/modules/microservices-tutorial/assets/attachments"
+
+function prepareAttachment() {
+   source_name=$1
+   zip ${tutorial_attachments}/${source_name}.zip ${tutorial_sources}/${source_name}
```
patriknw, line 8: when unzipping the folder structure should be without the number prefixes
```
shopping-cart-service/
shopping-analytics-service/
shopping-order-service/
```

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,21 @@
...
+}
+
+mkdir -p ${tutorial_attachments}
+
+prepareAttachment 00-shopping-cart-service-scala
+prepareAttachment 01-shopping-cart-service-scala
+prepareAttachment 02-shopping-cart-service-scala
+prepareAttachment 03-shopping-cart-service-scala
+prepareAttachment 04-shopping-cart-service-scala
+prepareAttachment 05-shopping-cart-service-scala
```
patriknw, line 18: 05 is the result of "Projection publishing to Kafka". That zip should also include the shopping-analytics-service

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,21 @@
...
+prepareAttachment shopping-cart-service-scala
+prepareAttachment shopping-analytics-service-scala
+prepareAttachment shopping-order-service-scala
```
patriknw, line 21: There should be one zip with the full example, i.e. including all 3
```
shopping-cart-service/
shopping-analytics-service/
shopping-order-service/
```

ignasi35:
Given the comments above, I think the approach I took here is wrong. Instead, I will create a temporal folder with the correct name (not prefixed, e.g. `01-`) where I'd copy the necessary contents (could be a mix of `05-shopping` and `analytics` for example) and then zip that temporal folder and move it to `assets/attachments`.

patriknw:
Sounds good. Then we can also remote the Antora tags there before zipping.

ignasi35:
This is ready for review.

I haven't worked on a solution to `remove the Antora tags there before zipping`. I am thinking something based on `sed` but I know it has some syntax differences between macOS and linux. I'm open to suggestions. I even think we can do that tag removal bit on a separate PR.

patriknw:
> I am thinking something based on sed but I know it has some syntax differences between macOS and linux.

Yes, whatever works. If it only works on Linux (Travis) I'm fine with that too.

> I even think we can do that tag removal bit on a separate PR.

Yes

ignasi35:
I've spiked the tag removal in https://travis-ci.com/github/akka/akka-platform-guide/builds/188417741#L226-L234 and it turned out to be simpler than I anticipated.

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,83 @@
...
+
+mkdir -p ${tutorial_attachments}
+
+## Remove Antora tags from codebase
+function removeTags() {
+   pushd ${tutorial_sources}
+   find . -type f -print0 | xargs -0 sed -i "s/\/\/ tag::[^\[]*\[.*\]//g" 
+   find . -type f -print0 | xargs -0 sed -i "s/# tag::[^\[]*\[.*\]//g" 
+   find . -type f -print0 | xargs -0 sed -i "s/\/\/ end::[^\[]*\[.*\]//g" 
+   find . -type f -print0 | xargs -0 sed -i "s/# end::[^\[]*\[.*\]//g"
```
patriknw, line 16: could we also remove `// <1>` callouts?

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+## Remove the tags used by Antora snippets from 
+## the codebase before zipping
+removeTags
+
+
+## gRPC service
+prepareTemporalFolder
+fetchProject 01-shopping-cart-service-scala shopping-cart-service
```
patriknw, line 54: I think we can create one for 00 also. g8 is the recommended seed for Scala but if that doesn't work for some reason they can download 00 when starting at the gRPC service.

patriknw (commented):
looking great

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,97 @@
...
+   ## remove tags from code
+   find . -type f -print0 | xargs -0 sed -i "s/\/\/ tag::[^\[]*\[.*\]//g" 
+   find . -type f -print0 | xargs -0 sed -i "s/\/\/ end::[^\[]*\[.*\]//g" 
+   
+   ## remove tags from config
+   find . -type f -print0 | xargs -0 sed -i "s/# tag::[^\[]*\[.*\]//g" 
+   find . -type f -print0 | xargs -0 sed -i "s/# end::[^\[]*\[.*\]//g" 
+
+   ## remove call-outs
+   find . -type f -print0 | xargs -0 sed -i "s/\/\/ <[0-9]*>//g"
```
ignasi35, line 23: This is the only regexp I didn't test.

scripts/prepare-dowloads.sh
octonato: Do we have any other reason for having a 00 variant? 

If the only reason is that .g8 may not work, then I don't think we should have it at all. 

That's yet another thing to keep in sync for us.

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,97 @@
...
+
+## Remove the tags used by Antora snippets from 
+## the codebase before zipping
+removeTags
+
+
+## empty codebase (this is a backup option for users 
+## that can't use the .g8 template)
+prepareTemporalFolder
+fetchProject 00-shopping-cart-service-scala shopping-cart-service
```
octonato, line 63: there is no `00-shopping-cart-service-scala`. @ignasi35, maybe you forgot to `git add .`?

scripts/prepare-dowloads.sh
```diff
@@ -0,0 +1,97 @@
...
+zipAndAttach step-2.zip
```
octonato, line 64: Maybe better if we don't number the zips and give it a meaningful name. Should that be easier for users to link back to tutorial step?

scripts/prepare-dowloads.sh
ignasi35: This PR must be merged after the one that splits the code.

There is no 04 and 05 either.  :-)

scripts/prepare-dowloads.sh
ignasi35: Sure thing.
Let's do that when adding the links to the assets (separate PR).

patriknw (approved):
LGTM

PR #242, opened 2020-10-05
Prefer local-preview

ignasi35:


README.md
```diff
@@ -10,8 +10,8 @@ Try it with:
 
 1. git clone https://github.com/akka/akka-platform-guide
 2. cd akka-platform-guide
-3. make html-author-mode
-4. open target/staging/snapshot/index.html
+3. make local-preview
+4. open http://localhost:8000
```
patriknw, line 14: What is the advantage?
This requires more local installation:
```
Access the documentation on http://localhost:8000
(cd target/staging/; python3 -m http.server)
/bin/sh: python3: command not found
make: *** [local-preview] Error 127
```

ignasi35: It's served over `http://` rather than `file:///`. I don't think it makes much of a difference, you are right.

PR #252, opened 2020-10-06
Remove antora tags

ignasi35:


ignasi35:
This was a hack to run an experiment in Travis.

PR #254, opened 2020-10-06
Akka links are local

### LINKED ISSUE #253: [style] Links to contents under akka.io are not external
We use the `tab-icon` double-link trick for any link that's not in the platform guide directly.

This is not 100% correct since we often link to contents under akka.io which should not be considered external either. For example:

```
{akka-management}[Akka Management] {akka-management}[{tab-icon}, window="tab"]
```

Should be:

```
{akka-management}[Akka Management]
```

ignasi35:
I used VSCode's global search and replace with ` \{akka[^{]*\{tab-icon[^\]]*\]` (note the initial blank space) and replaced it with `` (empty string).

Fixes #253


patriknw:
I'm not sure about this for two reasons:
1. Reference docs are different enough that we should have them as "external". It's easy to get lost when going to the reference docs and we want them to come back to the guide (or stay there as long as possible)
1. We don't know yet if the guide will be published at akka.io or lightbend.com yet (we are working on that decision)

ignasi35:
> I'm not sure about this for two reasons:
> 
>     1. Reference docs are different enough that we should have them as "external". It's easy to get lost when going to the reference docs and we want them to come back to the guide (or stay there as long as possible)

In that case, instead of manually making every single link include the tab-icon, I think we should make the theme automatically add it. If `xref` are used for in-guide links and regular links are used for everything else it would Just Work ™️.

>     2. We don't know yet if the guide will be published at akka.io or lightbend.com yet (we are working on that decision)

I thought that was decided already. My bad.

ignasi35:
I'll close this as it is becoming obsolete. The description has the regexp to apply the fix.

When we have a theme I would still try the theme-based approach.

Issue #247, opened 2020-10-06
[concepts] Reactive XYZ


ignasi35:
Reactive systems, reactive architecture, reactive programming, reactive streams, ... there's a lot of confusion so a page listing all of them explainig what each addresses and a paeg dedicated to reactive programming (and probaly a few of the others (Reactive Streams or Reactive Principles don't have a dedicated page at the moment) should help shed some light.

Issue #211, opened 2020-09-25
Introduce the reactive principles


ignasi35:
The section in `concepts` regarding properties of a reactive microsystem may include a small reference to https://principles.reactive.foundation/principles/index.html

Other concepts may also refer to https://principles.reactive.foundation/principles/index.html, e.g.: when introducing supervision, mention how that is part of embracing failure (principle III).

patriknw:
Links were added. Let's start with that and we can revisit later if we want more.

PR #255, opened 2020-10-06
Adds reactive disambiguation

ignasi35:
References #247 and #211

### DIFF
--- a/docs-source/docs/modules/concepts/nav.adoc
+++ b/docs-source/docs/modules/concepts/nav.adoc
@@ -15,6 +15,7 @@
 ** xref:akka-management.adoc[]
 
 * Patterns and techniques recommended for Reactive Microsystems
+** xref:reactive-disambiguation.adoc[]
 ** xref:event-sourcing.adoc[]
 ** xref:cqrs.adoc[]
 ** xref:message-driven-event-driven.adoc[Message Driven]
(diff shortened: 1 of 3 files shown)


octonato (commented):
I don't know if we should add this. Why do we need a disambiguation page at that stage? Where in the guide are we creating ambiguity that we need to solve? 

I feel that this is adding information on top of information, making the guide looks overwhelming. 

I don't think that making a distinction between Reactive Manifesto, Reactive Principles, Reactive System and Reactive Stream will help users understand how to build applications using Akka. I would prefer to have the guide more down to earth and directly covering clear use cases for Akka. 

If we believe its ambiguous, we should add such a page on the Reactive Foundation website, not here.

ignasi35:
Multiple concepts refer to the Reactive Principles. Several concepts are justified as "The preferred choice to stay reactive" and link to some term in the glossary of the reactive manifesto. Akka Streams reference the Reactive Streams.

Disambiguating Reactive Programming and Reactive systems is such an important topic that it even produced a whitepaper. The latest addition to the mix, the Reactive Principles and the Reactive Patterns, only made the mix worse. 

Having a page we can link to anytime we need to disambiguate will put a nail on the discussion.

patriknw (approved):
LGTM, but I agree with Renato's concern that the number of concepts are becoming scary and overwhelming. That can be solved by Ignasi's suggestion of not presenting all of them in the navigation list.

I think those are fundamental enough to add but let's select additional concepts carefully (Ignasi and I have talked about this).

ignasi35:
> LGTM, but I agree with Renato's concern that the number of concepts are becoming scary and overwhelming. That can be solved by Ignasi's suggestion of not presenting all of them in the navigation list.
> 
> I think those are fundamental enough to add but let's select additional concepts carefully (Ignasi and I have talked about this).

I have no intent on adding more concepts and I've already trimmed the list from the original plan drafted in #59.

Issue #249, opened 2020-10-06
[concepts] Akka modules


ignasi35:
The concepts section has scaffolding for all the akka modules used in the microservices tutorials:

- [x] akka core (actors) #260 
- [x] akka streams #244 
- [x] akka cluster #261 
- [x] akka persistence #262
- [x] akka projections #263 
- [x] akka management #234

patriknw:
Maybe we can steal some summaries from https://doc.akka.io/docs/akka/current/typed/guide/modules.html

ennru:
They are all filled in and merged now.

PR #262, opened 2020-10-07
Akka Persistence

ignasi35:
References #249

Stolen from https://doc.akka.io/docs/akka/current/typed/guide/modules.html#persistence

### DIFF
--- a/docs-source/docs/modules/concepts/pages/akka-persistence.adoc
+++ b/docs-source/docs/modules/concepts/pages/akka-persistence.adoc
@@ -6,5 +6,22 @@
 include::partial$include.adoc[]
 
 
+Just like objects in OOP, actors keep their state in volatile memory. When a system shuts down (gracefully or because of a crash) all data that was in memory is lost. Persistence provides patterns to enable actors to persist events that lead to their current state. Upon startup, events can be replayed to restore the state of the entity hosted by the actor. The event stream can be queried and fed into additional processing pipelines (an external Big Data cluster for example) or alternate views (like reports).
 
-ifdef::todo[TODO: this page is WIP]
+Persistence tackles the following challenges:
+
+* How to restore the state of an entity/actor when system restarts or crashes.
+* How to ensure reliable delivery of messages in face of network errors and system crashes.
+* How to introspect domain events that have led an entity to its current state.
+* How to leverage xref:event-sourcing.adoc[] in your application to support long-running processes while the project continues to evolve.
+* How to implement a xref:cqrs.adoc[CQRS] system.
+
+
+
+== Persistence Plugins
+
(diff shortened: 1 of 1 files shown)


docs-source/docs/modules/concepts/pages/akka-persistence.adoc
```diff
@@ -6,5 +6,22 @@
 include::partial$include.adoc[]
 
 
+Just like objects in OOP, actors keep their state in volatile memory. When a system shuts down (gracefully or because of a crash) all data that was in memory is lost. Persistence provides patterns to enable actors to persist events that lead to their current state. Upon startup, events can be replayed to restore the state of the entity hosted by the actor. The event stream can be queried and fed into additional processing pipelines (an external Big Data cluster for example) or alternate views (like reports).
```
octonato, line 9:
```suggestion
Just like objects in OOP, actors keep their state in volatile memory. When a system shuts down (gracefully or because of a crash) all data that was in memory is lost. Akka Persistence provides the means to persist an Actor state using Event Sourcing. Upon startup, events can be replayed to restore the state of the entity hosted by the actor. The event stream can be queried and fed into additional processing pipelines (an external Big Data cluster for example) or alternate views (like reports).
```

octonato (commented):
Good! 

Maybe worth mention on this intro that it's about Event Sourcing. We can then also improve the same in Akka reference. Let's refine it here.

ignasi35:
> Maybe worth mention on this intro that it's about Event Sourcing. ...

@octonato I've accepted your suggestion already. Any other improvement we should add?

docs-source/docs/modules/concepts/pages/akka-persistence.adoc
```diff
@@ -6,5 +6,22 @@
...
+
+
+
+== Persistence Plugins
+
+Storage backends for journals and snapshot stores are pluggable in the Akka persistence extension. Plugins maintained within the Akka organization include plugins for Cassandra and JDBC databases.
+
+NOTE: The xref:microservices-tutorial:index.adoc[Implementing Microservices with Akka tutorial] illustrates how to implement an Event Sourced CQRS application with Akka Persistence and the Akka Persistence Cassandra plugin.
+
+Learn more about Akka Persistence on the {akka}/typed/index-persistence.html[Akka docs (Persistence module)] {akka}/typed/index-persistence.html[{tab-icon}, window="tab"].
```
patriknw, line 27: "Learn more" section?

ignasi35: yeah, I'm doing them all.

octonato (approved):
Looks good. Nothing else to add, @ignasi35. 

I will merge it as Patrik's request for a "Learn more" section is addressed. Let's open another PR if needed.

PR #268, opened 2020-10-07
Initial Buildout of Lightbend Antora Theme with Lunr Search

rasummer:
@ennru 

This is the initial buildout of the latest Lightbend Antora skin that includes the Lunr Search integration. It uses a `supplemental_ui` directory that includes elements such as CSS, Images, JS and handlebars components. This is a newer way of extending the [default antora ui](https://docs.antora.org/antora-ui-default/templates/). 

For now I've been testing it with the `author-mode-site.yml` file locally. The idea is we no longer point to the Lightbend Theme `ui-bundle.zip` for the skin. We now point to Antora's stable default UI zip as found here: 

`https://gitlab.com/antora/antora-ui-default/-/jobs/artifacts/master/raw/build/ui-bundle.zip?job=bundle-stable` 

And then we extend that with the `supplemental_ui` files. Long term this should make it easier for me to keep us inline with new versions of the Antora Default UI and it's new features. It does mean that more of the UI is broken out and accessible here, so a little bit more maintenance for me when we update menus etc - as I'll need to do it in multiple repos - but I save time in other places, such as developing the `supplemental_ui`. 

It's working nicely, with the exception of the search feature. The Lunr search integration needs a bit more work to get it working with the set-up that we have in this repo as it's different to how I do it normally. I was hoping you'd be able to take a look and help figure it out. I think it has to do with the Makefile and the Docker set up. 

We need to have https://github.com/Mogztter/antora-lunr and https://github.com/Mogztter/antora-site-generator-lunr installed and then run:

`DOCSEARCH_ENABLED=true DOCSEARCH_ENGINE=lunr NODE_PATH="$(npm -g root)" antora --generator antora-site-generator-lunr author-mode-site.yml`

Please note that I've still got a few things on my list such as final menu/footer elements, and likely some CSS fixes, but I figured it would make sense to get something workable over into this repo as soon as possible so we can figure out how to get it running with the docker/makefile setup.

I'm happy to jump on a hangout online if it would help to figure this out in person. 

Rich


docs-source/author-mode-site.yml
```diff
@@ -25,6 +26,8 @@ asciidoc:
     review: ''
     todo: ''
     doc-title: 'Akka Samples'
+    page-pagination: ""
+    page-toctitle: "On This Page"
```
patriknw, line 30: looks like the "ON THIS PAGE" is now duplicated, both in the side bar and at the top of the page

docs-source/supplemental_ui/partials/supergroups.hbs
```diff
@@ -0,0 +1,85 @@
+<fieldset class="supergroup-switch">
```
ignasi35, line 1: I don't think we'll use all these groups. Maybe the scala-java and the sbt-maven (not even gradle).

Eventually, we maybe will need the EKS/GCP, but not now.

docs-source/supplemental_ui/partials/nav-menu.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="nav-panel-brand">
+	<img class="akka-logo akka-logo-platform" src="{{uiRootPath}}/img/akka-platform-full-color.svg">
+</div>
+{{#if env.DOCSEARCH_ENABLED}}
```
ignasi35, line 4: I don't see the search input. How can we enabled it (if possible at all) when using the local preview?

ignasi35 (approved):
I noticed we now have a `edit ths page` link on the top right corner but it doesn't work. Do we need more settings (root path of the repo)?

ignasi35:
@rasummer I noticed the font size in admonitions (TIP/WARNING/CAUTION/...) is smaller than the default text size. Is that on purpose? It looked a bit weird to me that admonitions are harder to read.

ignasi35:
@rasummer do you plan on adding the non-minified CSS and JS (and minification scripts)?

docs-source/supplemental_ui/partials/supergroups.hbs
rasummer: These were the existing groups from the current Lightbend theme. Feel free to remove or add groups as needed.

docs-source/author-mode-site.yml
rasummer: > @rasummer I noticed the font size in admonitions (TIP/WARNING/CAUTION/...) is smaller than the default text size. Is that on purpose? It looked a bit weird to me that admonitions are harder to read.

I still have some CSS work to complete. I will take a look at these.

rasummer: Yep 'On This Page' is now native to Antora, [https://docs.antora.org/antora/2.3/whats-new/#on-this-page-widget](https://docs.antora.org/antora/2.3/whats-new/#on-this-page-widget) .

We'll no longer need to add the following to each .adoc file:
```
:toc:
:toc-title: ON THIS PAGE
:toclevels: 2
```

If you need to turn off 'On This Page' on a per page basis, instructions are [here](https://docs.antora.org/antora/2.3/whats-new/#on-this-page-widget)

rasummer:
> @rasummer I noticed the font size in admonitions (TIP/WARNING/CAUTION/...) is smaller than the default text size. Is that on purpose? It looked a bit weird to me that admonitions are harder to read.

Will take a look.

rasummer:
> @rasummer do you plan on adding the non-minified CSS and JS (and minification scripts)?

I have a master repo where I have all the non-minified CSS and JS, as well as minification scripts. As we will be using this 'Supplemental UI' in other doc repos, I thought it made sense to keep the source files in one place and not have separate versions everywhere. Just bring over the final minified files to the doc repos.

This is the source repo: https://github.com/lightbend/antora-supplemental-ui-lightbend-theme

I would prefer to keep the source files in one place, but could be persuaded otherwise if you see a need. Happy to have others work in the master repo also.

docs-source/supplemental_ui/partials/nav-menu.hbs
rasummer: See the main PR comments at the top of the page that I addressed to Enno. There are a couple of things that need to be addressed to make Search work.

rasummer:
> I noticed we now have a `edit ths page` link on the top right corner but it doesn't work. Do we need more settings (root path of the repo)?

Yep, likely some settings, see: [https://docs.antora.org/antora/2.3/playbook/content-edit-url/ ](https://docs.antora.org/antora/2.3/playbook/content-edit-url/)

docs-source/author-mode-site.yml
rasummer: FYI, I've pulled out all of the .adoc in-page toc tags in the latest commit.

rasummer:
@ennru @patriknw @ignasi35 

FYI: If the cookie bar is getting annoying for you during local preview, simply comment out the JavaScript include

`<!--<script src="https://cdn.cookielaw.org/consent/ae5fff26-6829-4430-a8ae-e7efb10d58bf.js" type="text/javascript" charset="UTF-8"></script>-->`

in the following file: [docs-source/supplemental_ui/partials/head-scripts.hbs](docs-source/supplemental_ui/partials/head-scripts.hbs)

docs-source/supplemental_ui/partials/supergroups.hbs
rstento: Having the extra groups available isn't much overhead. I vote for leaving them because from previous experience, you never know when you might need them. Note that the defined tabs enable the selector at the top of the page. We can arbitrarily create tabs, it's just that the ones defined in the template enable that selector.

patriknw: By the way, I'd like to have Java first (first in the select list and to the left in the tabset).
I guess that is just a matter of defining it as `java-scala` instead of `scala-java`. We can do that ourselves here, but maybe something you want to standardize/recommend in the Lightbend theme.

patriknw: and then it should also be Maven/Gradle/sbt

rasummer: Updating the order sounds good to me, I will make the change in both repos and test.

ennru:
With #282 I've adapted the makefile to create an Antora Docker image with Lunr included which is than used to create the site including the search index.
Please rebase from main and see if it works for you as well.

ennru (approved):
Great, let's take it from here.
LGTM.

PR #271, opened 2020-10-08
Trying tabbed sources

patriknw:
Works as expected, but it would be nice if it could remember the choice when moving between pages. Akka docs even remember the choice when coming back to the site.

@rasummer any possibility that could be added?


docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -2,6 +2,7 @@
 :toc:
 :toc-title: ON THIS PAGE
 :toclevels: 2
+:page-supergroup-scala-java: Language
```
patriknw, line 5: This is how to define tabs. Other files are for the new theme.

docs-source/docs/modules/microservices-tutorial/pages/grpc-server.adoc
```diff
@@ -46,17 +47,40 @@ sbt compile
 
 Let's implement the `ShoppingCartService` trait. Create a `ShoppingCartServiceImpl` class that extends the generated `ShoppingCartService` trait. Implement the `addItem` method from the trait by logging the call and return a successful `Future`.
 
+[.tabset]
+Java::
++
+.src/main/java/shopping/cart/ShoppingCartServiceImpl.java
```
patriknw, line 53: I tried the filename labels, and it's included in the source code box. could be ok, but could be confusing to include the file name if they copy all of the code. (would have been nice with a copy icon, btw)

ignasi35: Agree, this looks confusing:

Screenshot 2020-10-08 at 16 31 05

Here's how it looks before using tabs:

Screenshot 2020-10-08 at 16 32 11

rasummer:
> Works as expected, but it would be nice if it could remember the choice when moving between pages. Akka docs even remember the choice when coming back to the site.
> 
> @rasummer any possibility that could be added?

It does store a cookie that remembers your choice so that when you navigate between pages it remembers your preferences. But to see this happening locally you'll need to run a server such as `python -m SimpleHTTPServer` so that the cookie can be stored against a domain such as `localhost`. If you are simply viewing the HTML page in your browser, the cookie doesn't get saved.

patriknw:
>  If you are simply viewing the HTML page in your browser, the cookie doesn't get saved.

ah, that was what I tried. thanks

Issue #181, opened 2020-09-10
Standardize footer


ignasi35:
Some pages use the footer:

```
 **This documentation was last published: {localdate}**
```

We should create a `partial` for it and include it everywhere.

patriknw:
wouldn't that be part of the style https://github.com/akka/akka-platform-guide/issues/3 ?

ignasi35:
> wouldn't that be part of the style #3 ?

Yeah, the theme can probably handle that.

ignasi35:
See also #268

patriknw:
I think this has been discussed and can be closed.

PR #275, opened 2020-10-09
Remove unnecessary info regarding last publish

ignasi35:
References #181

@rstento is this info something we want to keep around?

### DIFF
--- a/docs-source/docs/modules/ROOT/pages/index.adoc
+++ b/docs-source/docs/modules/ROOT/pages/index.adoc
@@ -36,4 +36,3 @@ Some development tasks arise from requirements on existing projects. The xref:ho
 
 include::how-to:partial$listing.adoc[]
 
-**This documentation was last published: {localdate}**


ignasi35:
> @rstento is this info something we want to keep around?

I mean, maybe adding the information regarding last publish is:

1. unnecessary
2. may be handled on the theme (see #268) so it's part of all pages (?)

rstento:
> > @rstento is this info something we want to keep around?
> 
> I mean, maybe adding the information regarding last publish is:
> 
> 1. unnecessary
> 2. may be handled on the theme (see #268) so it's part of all pages (?)

If it ends up being part of the theme it's good to remove. It is very handy to be able to verify that publication occurred at the time we expected it to.

rstento:
@ignasi35, I don't see it as part of the theme.  Few people are going to notice this, it is at the bottom of the page and only on one page, however, it is very useful for us as authors to be able to take a quick look and verify when the last changes were made and whether changes we just made have been published yet. I use this a lot.

ignasi35:
> @ignasi35, I don't see it as part of the theme. Few people are going to notice this, it is at the bottom of the page and only on one page, however, it is very useful for us as authors to be able to take a quick look and verify when the last changes were made and whether changes we just made have been published yet. I use this a lot.

My thinking was that we can add this information hidden in the HTML. By doing that on the theme it'd be available on any page. Then, as authors we can verify that info on every page (not even need to go find the info in a particular page). 

I understand your use-case and I'm fine closing this PR. May eventually come with an alternative but this is not something we should worry about now.

Thanks @rstento :-)

Issue #86, opened 2020-08-25
[EPIC] Port the Shopping sample to Java


ignasi35:
Should it demo Maven or Gradle or sbt or 2/3 or all?

- [x] port the shopping cart service #213 
- [x] port the orders service
- [x] port the analytics service

patriknw:
Not sbt.

Maven or Gradle? That could be a question for PM @jtownley.
It's not urgent because we will not start this until the dust has settled for the Scala version.

jtownley:
Have we a standard that is already in use? I would suggest we maintain
consistency as much as possible.

--James

ignasi35:
> Have we a standard that is already in use? I would suggest we maintain consistency as much as possible.

We don't have a standard, no. In general, sbt is always demoed and Maven/Gradle are demoed ad hoc. This sample, in particular, is trickier because it uses akka-gRPC which is plugin based so each implementation differs a bit. Demoing Gradle, for example, is not just adding dependencies, there's other plugin features to present (but there's the ref docs for that already). Akka-gRPC provides plugins for sbt, Maven and Gradle but maintaining all three becomes a bit tedious. Adding only support for either Gradle or Maven would be ideal from a maintenance PoV.

patriknw:
Since we will only use "standard" build features I think it will not be extremely much extra work to include both Maven and Gradle, but it adds up to things to change and test.

I suggest that we start with Maven and then add Gradle when things have stabilized (and that we see a clear demand for it).
I assume Maven is still more widely used than Gradle. If that is not the case we can do the opposite.

patriknw:
BTW Lagom only has support for Maven so that is another reason for starting with Maven.

jtownley:
Seems like maven is the choice then :)


--James

PR #277, opened 2020-10-09
Java order and analytics services, #86

patriknw:
References #86


ignasi35 (approved):
LGTM

I like that default ports are now set to the same values in all services (and only local runs use customized ports)

patriknw:
> I like that default ports are now set to the same values in all services (and only local runs use customized ports)

yeah, that must have been an oversight

patriknw:
Don't know why Travis got lost here, but it's all green at https://travis-ci.com/github/akka/akka-platform-guide/builds/189039257 (I just rebuilt it)

PR #290, opened 2020-10-13
Renumber tutorial; remove previous/next links

ennru:
* Correct tutorial numbering (Roman numbers)
* Remove the previous/next links as they are part of the theme now

References #268


patriknw (commented):
This might be in conflict with @rstento work in progress? Then it might be better to wait.

ennru:
> This might be in conflict...

These are rather tiny changes. No files are renamed or such.

ignasi35 (approved):
LGTM

PR #297, opened 2020-10-14
Lightbend links in toolbar and footer.

rstento:
This changes the menu item from Documentation, which is a bit confusing since the reader is looking at a documentation page to Library, which implies a list of other resources. It also aligns these links along the current LB strategy. @rasummer, for some of the shorter names in the toolbar menu, I had to add a <br> because they would fall on the same line as another entry. I assume you could fix this with css instead? 

@patriknw, do we expect this guide to be versioned? If not, we should change the version number in the antora file to "master", this would prevent the "Snapshot" version selector from appearing at the bottom of the TOC.

### DIFF
--- a/docs-source/supplemental_ui/partials/footer-nav.hbs
+++ b/docs-source/supplemental_ui/partials/footer-nav.hbs
@@ -1,25 +1,24 @@
 <nav>
 	<section>
-		<h3>Lightbend Documentation</h3>
+		<h3>Lightbend Docs</h3>
 		<ul>
-				<li><a class="navbar-item" href="https://akka.io/docs/">Akka</a></li>
-				<li><a class="navbar-item" href="https://doc.akka.io/docs/alpakka/current/">Alpakka</a></li>
-				<li><a href="https://doc.akka.io/docs/akka-enhancements/current/akka-resilience-enhancements.html">Akka Resilience Enhancements</a></li>
-				<li><a href="https://doc.akka.io/docs/akka-enhancements/current/akka-persistence-enhancements.html">Akka Persistence Enhancements</a></li>
-				<li><a href="https://cloudflow.io/docs/current/">Cloudflow</a></li>
-				<li><a href="https://developer.lightbend.com/docs/cloudflow/current/">Cloudflow Enterprise Features</a></li>
-				<li><a href="https://developer.lightbend.com/docs/fortify/current/">Fortify SCA for Scala</a></li>		
+				<li><a href="https://developer.lightbend.com/docs/cloudflow/current/">Akka Data Pipelines</a></li>
+				<li><a href="https://docs.cloudstate.com/">Akka Serverless</a></li>
+				<li><a href="https://developer.lightbend.com/docs/fortify/current/">Fortify SCA for Scala</a></li>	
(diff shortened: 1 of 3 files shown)


rasummer:
@rstento LGTM, I will fix the need to add `<br>` tags to the footer with a css solution. Adding an issue over in the repo where I develop the frontend: https://github.com/lightbend/antora-supplemental-ui-lightbend-theme/issues/2

octonato:
@rstento, we need to think about the versioning. I do think that we can skip it for now, at least. 

I don't know if we want to associate a version with it. The guide covers Akka, but also some associated libraries that have their own release cycle, like akka-projections and akka-grpc. So, we can't pick a version number that is intuitive for the users. 

We can say, the guide is version 1.0, but what that even mean for the users? We can't use Akka's own version that won't make sense either. 

The reference documentations have a version for each released version, but the guide assembles multiple versions. I think we should treat it differently. 

What we may need to do at some point is to explicitly enumerate the versions (akka, akka-grpc, etc) used in the guide. If someone is on previous version of any library, they can use the guide as a source of high-level information about our stack, but they will need to consult the reference documentation for the version their are using.

patriknw:
I agree with what Renato said about versioning. Let's start without versions. Later we might have to support at least one old version branch, maybe corresponding to minor version of Akka core (2.6.x vs 2.7.x).

docs-source/supplemental_ui/partials/footer-nav.hbs
```diff
@@ -1,25 +1,24 @@
...
-				<li><a class="navbar-item" href="https://www.scala-sbt.org/documentation.html">sbt</a></li>
-				<li><a class="navbar-item" href="https://docs.scala-lang.org/">Scala</a></li>
-				<li><a class="navbar-item" href="http://slick.lightbend.com/doc/3.2.1/">Slick</a></li>
-				<li><a href="https://developer.lightbend.com/docs/telemetry/current/home.html">Lightbend Telemetry</a></li>			
+				<li><a href="https://akka.io/docs/">Akka</a></li>
+        <li><a href="https://doc.akka.io/docs/alpakka/current/">Alpakka</a></li>
+				<li><a href="https://cloudflow.io/docs/current/">Cloudflow</a></li>	
+				<li><a href="https://cloudstate.io/docs/">Cloudstate</a></li>	
+				<li><a href="https://www.scala-sbt.org/1.x/docs/">sbt</a></li>
+				<li><a href="https://docs.scala-lang.org/">Scala</a></li
```
ennru, line 20:
```suggestion
				<li><a href="https://docs.scala-lang.org/">Scala</a></li>
```

docs-source/supplemental_ui/partials/header-content.hbs
```diff
@@ -51,7 +57,7 @@
                   <a href="https://www.lightbend.com/customers">Who Uses Lightbend?</a>
                   <a href="https://www.lightbend.com/lightbend-subscription">Why Get A Subscription?</a>
                   <a href="https://www.lightbend.com/company/news">News & Press</a>
```
ennru, line 59:
```suggestion
                  <a href="https://www.lightbend.com/company/news">News &amp; Press</a>
```

ennru (approved):
LGTM.
I'll PR the change to become versionless separately as it will affect some directories.

PR #298, opened 2020-10-15
No versioning of the guide

ennru:
Antora uses the magic version name `master` to be used without the version in the URL, which affects the version selector in the lower left corner. Annoying that we can't use "main" here as we do for the main branch.

* update directories to not contain "snapshot"
* Let the default make target will use the local version of the sources and not pull it via git anymore.
* Use the new theme even in `site.yml`

References #297

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ html: clean docker-image
 		-t ${antora_docker_image}:${antora_docker_image_tag} \
 		--cache-dir=./.cache/antora \
 		docs-source/site.yml
-	@echo "Done file://${target_dir}/snapshot/index.html"
+	@echo "Done ${target_dir}/index.html"
 	@echo "# Tech Hub wants to know what to publish:"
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -32,9 +32,9 @@ html: clean docker-image
 		-t ${antora_docker_image}:${antora_docker_image_tag} \
 		--cache-dir=./.cache/antora \
 		docs-source/site.yml
-	@echo "Done file://${target_dir}/snapshot/index.html"
+	@echo "Done ${target_dir}/index.html"
```
octonato, line 35: The reason we have `file://` is to be able to click on it in the Termimal and open it in the browser

octonato (approved):
LGTM, but maybe we should reconsider the removal of `file://`

Makefile
ennru: Yes, that's why I added it, but it works better without.

Issue #265, opened 2020-10-07
Detect broken links


ignasi35:
We should have an automated mechanism to detect broken links.

Here's a mistake I make often: I declare `:foo:  http://com.com/` and use `{foo}/subpath` which renders `http://com.com//path`.

ennru:
The link checker `make check-link` now works. Let's create a scheduled task to verify links.

Refs #283

ennru:
This is what my home-grown link checker thinks about the current status in `main`:

## Non-HTTP OK pages
https://developer.lightbend.com/docs/akka-platform-guide status 404 Not Found
 - concepts/akka-actor.html
 - concepts/akka-management.html
 - ... 41 more

https://doc.akka.io/docs/akka/current/split-brain-resolver status 404 Not Found
 - concepts/akka-management.html

https://www.reactivemanifesto.org//glossary status 404 Not Found
 - concepts/reactive-streams.html

https://www.amazon.com/Domain-Driven-Design-Distilled-Vaughn-Vernon-ebook/dp/B01JJSGE5S/ref=sr_1_1?dchild=1&keywords=Domain-Driven+Design+Distilled&qid=1599820686&s=books&sr=1-1 status 405 Method Not Allowed
 - concepts/ddd.html

https://www.amazon.com/Domain-Driven-Design-Tackling-Complexity-Software/dp/0321125215/ref=sr_1_1?dchild=1&keywords=Domain-Driven+Design&qid=1599820577&s=books&sr=1-1 status 405 Method Not Allowed
 - concepts/ddd.html


## Redirected URLs
https://principles.reactive.foundation/principles should be
http://principles.reactive.foundation/principles/
 - concepts/isolation.html

https://portal.lightbend.com/ should be
/account/login
 - concepts/akka-actor.html
 - concepts/akka-management.html
 - ... 41 more

ignasi35:
```
## Redirected URLs
https://principles.reactive.foundation/principles should be
http://principles.reactive.foundation/principles/
```

Hmm, I get the trailing `/` but default to http?

ennru:
> Hmm, I get the trailing / but default to http?

My browser doesn't change to http. Must be something in my tool.

ignasi35:
> My browser doesn't change to http. Must be something in my tool.

It's a double redirect:

```
$ curl -v https://principles.reactive.foundation/principles
*   Trying 216.239.38.21...
* TCP_NODELAY set
...
[...]

PR #299, opened 2020-10-15
Run the link checker (and prepare for a scheduled build with it)

ennru:
References #265

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,13 @@ services:
 git:
   depth: 500
 
-before_install:
-  # using jabba for custom jdk management
-  - if [ ! -f ~/.jabba/jabba.sh ]; then curl -L -v --retry 5 -o jabba-install.sh https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh && bash jabba-install.sh; fi
-  - . ~/.jabba/jabba.sh
-  - jabba install ${JDK:=adopt@~1.11-0}
-  - jabba use ${JDK:=adopt@~1.11-0}
-
 jobs:
   include:
     - stage: test
       script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -96,7 +100,7 @@ jobs:
 
 stages:
   - name: test
-    if: NOT tag =~ ^v
+    if: (NOT tag =~ ^v) AND (type != cron)
```
ignasi35, line 103: What's the purpose of disabling tests in cron builds? 

If a test fails it prevents the publishing, so the question is whether that's a valid failure or a flaky test... hmmm.

ennru: I don't think it is useful to run these tests when there are no changes and it spares us some CI cycles for other repos.

ignasi35:
this build has a dirty cache in travis. You can either rebase over main once #313 is merged, or close the PR and create a new one.

.travis.yml
ignasi35: OFC! This is not using snapshot artifacts from upstream. 👍🏼

.travis.yml
```diff
@@ -6,19 +6,13 @@ services:
 git:
   depth: 500
 
-before_install:
-  # using jabba for custom jdk management
-  - if [ ! -f ~/.jabba/jabba.sh ]; then curl -L -v --retry 5 -o jabba-install.sh https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh && bash jabba-install.sh; fi
-  - . ~/.jabba/jabba.sh
-  - jabba install ${JDK:=adopt@~1.11-0}
-  - jabba use ${JDK:=adopt@~1.11-0}
```
ignasi35, line 14: Where did this go? Do you suggest we rely on the JVM installed by default in the `language: scala` flavor of TravisCI VMs?

PR #300, opened 2020-10-15
build: try to use official adoptopenjdk package for debian instead of jabba

jrudolph:


patriknw (approved):
LGTM

ennru (changes requested):
Please don't opt for a special solution in this repo.
We need to understand what happens and switch to a proper solution.

ennru:
As the cache cleaning hack was successful, I'll close this.

jrudolph:
Using the official package is far less special than using jabba or any other third party installer. :) After all it's probably jabba which got the cache into the broken state...

PR #302, opened 2020-10-15
Fix syntax error

ignasi35:
see the error in the red link (`concepts`)  on https://developer.lightbend.com/docs/akka-platform-guide/#:concepts:index.adoc


ennru (approved):
LGTM.

Issue #258, opened 2020-10-07
Include Java code in tutorial


patriknw:
include snippets to be shown as separate Scala/Java tabs

Some text content should also be adjusted based on the choice. I don't know if there is a solution for this yet or if we have to develop some Antora addition?

johanandren:
I've tried to keep the include tags to correspond with the Scala ones in #213 but it will surely need work through to check that it is ok.

patriknw:
I never started this. Open for grabs.

Issue #257, opened 2020-10-07
Split up the Java example code in parts


patriknw:
Corresponding to #199

johanandren:
Depends on PR #213

PR #304, opened 2020-10-15
[Java] step-5: projecting to a gRPC service

octonato:
References #257 and #258 

I'm going backwards with the PR so on each step I can remove a little bit of code until we get to the equivalent of a Java seed template.

- [x] need to add travis build for copied java example
- [x] script to copy identical files

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
@@ -17,8 +17,8 @@ ifdef::todo[TODO link to synchronous communication concepts]
 
 .Available .zip downloads:
 ****
-* link:_attachments/5-shopping-cart-projection-kafka-scala.zip[Source] that includes all previous tutorial steps and allows you to start with the steps on this page.
-* link:_attachments/6-shopping-cart-complete-scala.zip[Source] with the steps on this page completed.
+* [.group-scala]#link:_attachments/5-shopping-cart-projection-kafka-scala.zip[Source]# [.group-java]#link:_attachments/5-shopping-cart-projection-kafka-java.zip[Source]# that includes all previous tutorial steps and allows you to start with the steps on this page.
+* [.group-scala]#link:_attachments/6-shopping-cart-complete-scala.zip[Source]# [.group-java]#link:_attachments/6-shopping-cart-complete-java.zip[Source]# with the steps on this page completed.
 ****
 
 == Another gRPC service
@@ -38,6 +38,13 @@ include::example$shopping-order-service-scala/src/main/protobuf/ShoppingOrderSer
 
 Generate code by compiling the project:
 
+[.group-java]
+[source,shell script]
+----
+mvn compile
+----
+
+[.group-scala]
 [source,shell script]
 ----
 sbt compile
@@ -50,7 +57,7 @@ Java::
 +
 [source,java]
 ----
-TODO
+include::example$shopping-order-service-java/src/main/java/shopping/order/ShoppingOrderServiceImpl.java[]
 ----
 
 Scala::
@@ -67,7 +74,7 @@ Java::
 +
 [source,java]
 ----
-TODO
+include::example$shopping-order-service-java/src/main/java/shopping/order/ShoppingOrderServer.java[]
 ----
 
 Scala::
@@ -84,7 +91,7 @@ Java::
 +
 [source,java]
 ----
-TODO
+include::example$shopping-order-service-java/src/main/java/shopping/order/Main.java[]
 ----
 
 Scala::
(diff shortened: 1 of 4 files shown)


scripts/prepare-downloads.sh
```diff
@@ -95,3 +95,43 @@ fetchProject shopping-cart-service-scala shopping-cart-service
...
+
+## complete entity
+# prepareTemporalFolder
+# fetchProject 03-shopping-cart-service-java shopping-cart-service
+# zipAndAttach 3-shopping-cart-event-sourced-complete-java.zip
+
+## projection query
+# prepareTemporalFolder
+# fetchProject 04-shopping-cart-service-java shopping-cart-service
+# zipAndAttach 4-shopping-cart-projection-java.zip
```
octonato, line 124: I will base next PRs on this one. Actually, each step will be based on it's next step. As we move I will be uncommenting the zips.

octonato (commented):
There is still something incomplete on this PR. 

On these paragrahps, we talk about creating another project using the template. 

https://github.com/akka/akka-platform-guide/pull/304/files#diff-512fd5e58856e6355e329fdcdcdcab35ad8acd66810a19767df0f28648b40979R28

We don't have a see template for java. We may need to create a seed for the order service and link it here. Same for the analytics. 

I will open an issue for this.

scripts/prepare-downloads.sh
```diff
@@ -95,3 +95,43 @@ fetchProject shopping-cart-service-scala shopping-cart-service
...
+
+## projection kafka
+prepareTemporalFolder
+fetchProject 05-shopping-cart-service-java shopping-cart-service
+fetchProject shopping-analytics-service-java shopping-analytics-service
+zipAndAttach 5-shopping-cart-projection-kafka-java.zip
```
patriknw, line 130: the reason I have been suggesting 05 is that if we every use more than 9 parts the file name order is better with two digits, but we can change if that ever becomes an issue

patriknw (commented):
I think I'll wait with review until you have updated the copy-identical script. Surprised that you didn't start there. I assume most identical files will correspond to the Scala example.

scripts/prepare-downloads.sh
octonato: That was my though, add a 0 if needed. 
I was not expecting it to grow in more step either, but we never know.

octonato:
> I think I'll wait with review until you have updated the copy-identical script. Surprised that you didn't start there. I assume most identical files will correspond to the Scala example.

I completely forget about the copy-identical script when I started, but I find it easier to take the full example and remove what is added in step 5. 

On step 4, I will take the completed step 5 and remove code from step 4. That gives me better control I think.

patriknw:
Working backwards is the right way. That was what I did for Scala and that is how the copy script is structured. Well, it copies some from seed 00 also.

docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
```diff
@@ -130,9 +141,9 @@ include::example$shopping-cart-service-scala/src/main/scala/shopping/cart/SendOr
 
 <3> call the `ShoppingOrderService`
 
-In the order we need to include the list of items and their quantities. That information is not included in the `ChoppingCart.CheckedOut` event, but we can retrieve it by asking the `ShoppingCart` entity for it.
```
ignasi35, line 133: 🎻

docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
```diff
@@ -130,9 +141,9 @@ include::example$shopping-cart-service-scala/src/main/scala/shopping/cart/SendOr
...
+In the order we need to include the list of items and their quantities. That information is not included in the `ShoppingCart.CheckedOut` event, but we can retrieve it by asking the `ShoppingCart` entity for it.
 
-If the call to the `ShoppingOrderService` fails the returned `Future[Done]` is failed and the Projection is automatically restarted from the previously saved offset. This will result in that the call to the `ShoppingOrderService` will be retried. Since the Projection has `at-least-once` semantics the `ShoppingOrderService` must be idempotent, i.e. handle duplicate order attempts for the same `cartId`.
+If the call to the `ShoppingOrderService` fails the returned [.group-scala]#`Future[Done]`# [.group-java]#`CompletionStage[Done]`# is failed and the Projection is automatically restarted from the previously saved offset. This will result in that the call to the `ShoppingOrderService` will be retried. Since the Projection has `at-least-once` semantics the `ShoppingOrderService` must be idempotent, i.e. handle duplicate order attempts for the same `cartId`.
```
ignasi35, line 146:
```suggestion
If the call to the `ShoppingOrderService` fails the returned [.group-scala]#`Future[Done]`# [.group-java]#`CompletionStage<Done>`# is failed and the Projection is automatically restarted from the previously saved offset. This will result in that the call to the `ShoppingOrderService` will be retried. Since the Projection has `at-least-once` semantics the `ShoppingOrderService` must be idempotent, i.e. handle duplicate order attempts for the same `cartId`.
```

Let's fix this on a separate PR. I want to merge this one now.

ignasi35 (approved):
@octonato there's a silly typo

PR #308, opened 2020-10-16
Move to sdkman and away from travis provided sbt

ignasi35:
Alternative to #308 

References https://github.com/akka/akka-meta/issues/141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
-language: scala
+version: ~> 1.0 # needed for imports
+
+import: scala/scala-dev:travis/default.yml
+
+language: java
 services:
   - docker
 
@@ -6,12 +10,14 @@ services:
 git:
   depth: 500
 
+
+env:
+  global:
+    - ADOPTOPENJDK=11
 before_install:
-  # using jabba for custom jdk management
-  - if [ ! -f ~/.jabba/jabba.sh ]; then curl -L -v --retry 5 -o jabba-install.sh https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh && bash jabba-install.sh; fi
-  - . ~/.jabba/jabba.sh
-  - jabba install ${JDK:=adopt@~1.11-0}
-  - jabba use ${JDK:=adopt@~1.11-0}
+  # using sdkman for SDKs management
+  - sdk install sbt 1.3.13
+  - sdk use sbt 1.3.13
 
 jobs:
   include:
@@ -24,7 +30,7 @@ jobs:
     - stage: test
       script:
         - cd docs-source/docs/modules/microservices-tutorial/examples/shopping-analytics-service-scala
-        - sbt -jvm-opts .jvmopts-travis "test; scalafmtCheckAll"
+        - sbt "test; scalafmtCheckAll"
       name: "Shopping analytics service (Scala)"
     - script:
         - cd docs-source/docs/modules/microservices-tutorial/examples/shopping-analytics-service-java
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,12 +10,14 @@ services:
...
+    - ADOPTOPENJDK=11
 before_install:
-  # using jabba for custom jdk management
-  - if [ ! -f ~/.jabba/jabba.sh ]; then curl -L -v --retry 5 -o jabba-install.sh https://raw.githubusercontent.com/shyiko/jabba/0.11.2/install.sh && bash jabba-install.sh; fi
-  - . ~/.jabba/jabba.sh
-  - jabba install ${JDK:=adopt@~1.11-0}
-  - jabba use ${JDK:=adopt@~1.11-0}
+  # using sdkman for SDKs management
+  - sdk install sbt 1.3.13
+  - sdk use sbt 1.3.13
```
ignasi35, line 20: Use `sbt` provided by sdkman, because why not.

.travis.yml
```diff
@@ -106,18 +112,10 @@ stages:
...
-  - find $HOME/.sbt -name "*.lock" -delete
-
 cache:
   directories:
    - $HOME/.m2
    - $HOME/.ivy2/cache
-   - $HOME/.sbt/boot
-   - $HOME/.sbt/launchers
-   - $HOME/.cache/coursier
-   - $HOME/.jabba
```
ignasi35, line 119: All the caching management is inherited from the parent build.

ignasi35:
Something's weird:

```
sbt -jvm-opts .jvmopts-travis "test; scalafmtCheckAll"
[info] welcome to sbt 1.3.13 (AdoptOpenJDK Java 1.8.0_265)
...
[info] set current project to shopping-cart-service (in build file:/Users/ignasi/git/github/akka/akka-platform-guide/main/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/)
[error] Expected ';'
[error] -jvm-opts
```

.travis.yml
ignasi35: This is vanilla `sbt` instead of the Travis-provided `sbt-extras`. As a consequence, we can no longer use `-jvm-opts` but we were not using them anyway:

 - when we invoked `sbt -jvm-opts .jvmopts-travis` we had already `cd`'d into a folder with no `.jvmopts-travis`

.travis.yml
```diff
@@ -1,24 +1,32 @@
...
 
 jobs:
   include:
     - stage: test
       script:
         - cd docs-source/docs/modules/microservices-tutorial/examples/shopping-analytics-service-scala
-        - sbt -jvm-opts .jvmopts-travis "test; scalafmtCheckAll"
+        - ls -la 
+        - pwd
+        - sbt "test; scalafmtCheckAll"
```
ignasi35, line 29: We can eventually remove the `ls -la` and the `pwd` but they were useful to debug.

.travis.yml
```diff
@@ -1,24 +1,32 @@
-language: scala
+version: ~> 1.0 # needed for imports
+
+import: scala/scala-dev:travis/default.yml
```
ignasi35, line 3: Use `scala-dev`'s parent build.

.travis.yml
```diff
@@ -1,24 +1,32 @@
...
+
+language: java
```
ignasi35, line 5: Using `java` instead of `scala` so not `sbt` is installed out of the box (we'll use the one installed by sdkman)

ignasi35: Note: changing the language also had the nice side-effect that the polluted cache was no longer visible because travis caches by language/branch/etc...

ennru (changes requested):
Let's hold this off until we agree on a general direction for CI.

ignasi35:
@ennru even if we don't merge this PR, we should keep in mind these diffs:

```diff
- sbt -jvm-opts .jvmopts-travis "test; scalafmtCheckAll"
+ sbt "test; scalafmtCheckAll"
```

which I sneaked into this PR.

It doesn't matter if we stay on travis or move to some other CI provider, we should no longer use `-jvm-opts .jvmopts-travis` in the jobs.

ennru:
I added it (in many repos) to make output colouring work. But that breaks with sbt 1.4.0 again anyway...

ignasi35:
I think we can close instead of keeping it as DRAFT. We can restore it if/when we think it's the right move.

PR #311, opened 2020-10-16
Don't cache JDKs

ignasi35:
See https://docs.travis-ci.com/user/caching#things-not-to-cache


ignasi35:
This was an experiment rather than a PR

PR #312, opened 2020-10-16
Remove unused argument jvm-opts

ignasi35:
When `cd`-ing into each project, there is no `.jvmopts-travis` file and, still, `sbt` in travis works as expected. I think `sbt-extras` (which is the one provided by TravisCI) doesn't fail when `-jvm-opts` points to a non-existing file.


ignasi35:
Done in #308

PR #319, opened 2020-10-16
Run the link checker

ennru:
Replaces #299


ennru:
The Link check run quite quickly, I guess we can keep it in every build and decide later if we'd want a cron job with it.

patriknw (approved):
LGTM

Issue #317, opened 2020-10-16
Broken style


octonato:
Something went wrong recently. Styling is broken for this section. I remember that this was displaying correctly. 

https://developer.lightbend.com/docs/akka-platform-guide/microservices-tutorial/entity.html#_run

![CleanShot 2020-10-16 at 14 07 26](https://user-images.githubusercontent.com/502982/96256684-66067e00-0fb9-11eb-90ed-2698c053cd44.png)

@rasummer, I'm not sure if that is due to theme changes or because we are using `+` to indent the content.

rasummer:
The CSS changes I made yesterday were mostly limited to some background colors and styling the sidebar. Assuming all content needs to be indented under `#2` and using the `+` symbol to keep it all together? I'll take a look today.

octonato:
Yes, that's what I think it was the intention. All indented under #2. 

It may be some unforeseen combination that is causing it. The original commit, that was displaying correctly, didn't have the indentation.

rasummer:
My guess is it's a Asciidoc syntax issue, I will take a look regardless.

PR #321, opened 2020-10-16
Fixes indent issue

rasummer:
References #317 

@octonato I think the issue was having a `+` symbol on each side of the comment. When I remove the `+` above the comment the ident looks good, see below. 

fix-indent


octonato (approved):
Oh! Great. Thanks @rasummer

PR #322, opened 2020-10-16
[Java] step-3: projection item popularity

octonato:


docs-source/docs/modules/how-to/pages/from-crud-to-eventsourcing.adoc
```diff
@@ -57,7 +57,7 @@ In Event Sourcing, on the other hand, there is a single table (referred to as _J
 
 A (very) simplified version of a `ShoppingCart` can be modelled using two commands, and two events:
 
-[source,scala]
+[source,scala,indent=0]
```
patriknw, line 60: what is this? why do we have to add that everywhere? seems like a style thing?

patriknw (approved):
LGTM, with an Antora question

docs-source/docs/modules/how-to/pages/from-crud-to-eventsourcing.adoc
octonato: The indentation is preserved in Asciidoc. This wasn't obvious with Scala code because we use tab=2. And it also only becomes more clear when including only fragments (not whole file), like in this example:

CleanShot 2020-10-20 at 10 24 15

Here is the same, but in Scala:

CleanShot 2020-10-20 at 10 26 31

I have noticed the messy indentation before and I could have added only where it's needed, but to save time, I did a find-and-replace. Then I'm sure that we always have the right indent doesn't matter the content. 

We could indeed solved that with styling, but we can use it like that for now until we solved via styling.

octonato: The above example is not the best one. And it reveals another issue. 

I tried it out, and the above fragment won't indent correctly. The reason is that it's using different fragments on the same file and toggling fragments. Asciidoc will in that case preserve the indentation. 

Here is an example where the indentation gets fixed: 

**Before**

CleanShot 2020-10-20 at 10 40 53

**After**

CleanShot 2020-10-20 at 10 41 00

patriknw: ok, would have been good with some global default for that, but doesn't matter

PR #325, opened 2020-10-19
Drop caches on demand (see #313)

ennru:
Introduce an environment flag to control if the caches should be flushed.

References #313

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,10 @@ before_install:
   - jabba install ${JDK:=adopt@~1.11-0}
   - jabba use ${JDK:=adopt@~1.11-0}
 
+env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,10 @@ before_install:
   - jabba install ${JDK:=adopt@~1.11-0}
   - jabba use ${JDK:=adopt@~1.11-0}
 
+env:
+  global:
+    - FLUSH=false
```
ignasi35, line 18:
```suggestion
    - DELETE_CACHE=false
```

(pet peeve) I think `FLUSH` is overloaded.

ignasi35 (approved):

PR #334, opened 2020-10-20
proofreading of concepts

patriknw:
Nothing major, only many small things


docs-source/docs/modules/concepts/pages/deployment-and-container-orchestration.adoc
```diff
@@ -3,7 +3,7 @@
...
+To deploy an application built using Akka Cluster you can follow the {akka-management}/kubernetes-deployment/index.html[guide in the Akka Management documentation] {akka-management}/kubernetes-deployment/index.html[{tab-icon}, window="tab"]. That guide will help you complete your code and configure the application. It will also show you how to create a <<Docker>> container for your application and how to create the <<Kubernetes>> resources to successfully deploy the application. Finally, it will explain which are the relevant settings you need to consider when deploying an Akka Cluster application on a Kubernetes cluster.
```
ignasi35, line 6: Hmmm, I was trying to link the anchor-id (`#docker`)  instead of the section title (`Docker`) but I used the wrong syntax but I think that was wrong.

We can keep it with `<<Docker>>` and, if it ever breaks, we'll migrate to linking to the anchor-id

ignasi35 (approved):
Quite the cleanup. thanks @patriknw

PR #335, opened 2020-10-20
[Java] step-2: complete shopping cart entity

octonato:


docs-source/docs/modules/microservices-tutorial/pages/complete-entity.adoc
```diff
@@ -112,14 +112,14 @@ Scala::
 include::example$03-shopping-cart-service-scala/src/test/scala/shopping/cart/ShoppingCartSpec.scala[tag=checkout]
 ----
 
-Commands should be handled differently when the cart has been checked out. `AddItem` is no longer allowed after checkout. Therefore we refactor the `handleCommand` method to two separate methods `openShoppingCart` and `checkedOutShoppingCart` that are used depending on the `checkedOut` state. The previous code for `AddItem` goes into the `openShoppingCart` method.
+Commands should be handled differently when the cart has been checked out. `AddItem` is no longer allowed after checkout. [.group-scala]#Therefore, we refactor the `handleCommand` method to two separate methods `openShoppingCart` and `checkedOutShoppingCart` that are used depending on the `checkedOut` state. The previous code for `AddItem` goes into the `openShoppingCart` method as well as the new `Checkout` command.# [.group-java]#Therefore, we refactor the `commandHandler` method to define different handlers depending on the `checkedOut` state. `AddItem` and `Checkout` are only handled when the cart is open (not checked-out).#
```
octonato, line 115: It turns out that the impl in Java and Scala are quite different when it comes to state dependent handlers. Splitting the java one in two separated builders seemed overkill in that case, so instead I opt to keep it them all together on the same builder.

docs-source/docs/modules/microservices-tutorial/pages/complete-entity.adoc
```diff
@@ -28,7 +28,7 @@ Add the following commands to the `ShoppingCart`:
 * (Optional) `RemoveItem` - remove an item from the cart
 * (Optional) `AdjustItemQuantity` - adjust the quantity of an item in the cart
 
-If you want to reduce the scope of the exercise you can skip `RemoveItem` and `AdjustItemQuantity`. They are not mandatory for subsequent steps of the tutorial.
+If you want to reduce the scope of the exercise you can skip `RemoveItem` and `AdjustItemQuantity`. They are not mandatory for subsequent steps of the tutorial and its details won't be covered on this page. This is intentionally left out as an optional exercise.
```
octonato, line 31: Previously we were showing the optional command and event handlers as snippets in the page. I remove it because we don't really cover them and it becomes a distraction. 

Instead, I'm suggesting that we explicitly leave it out as an exercise for users. Later on, we mention that the full impl for this step contains the extras. People can have look if they want  or try to implement on their own.

docs-source/docs/modules/microservices-tutorial/pages/complete-entity.adoc
octonato: I'm reviewing this. I'm not happy with the results when showing it on screen. I think we should make then as close as possible. So, maybe not that overkill to split the java builders.

octonato: Did a push to add split builders for java. 

If we agree that's what we want, we need to adapt the code in step 4, 5 and final one.

rstento: @octonato, glad you did this. Using the inline .group-scala and .group-java assumes that the user will have selected what they want to see previously, either at the top of the page or in a tabbed example block. If they haven't, and for example are interested in scala, but java is the default, they will only see the java info and may not realize that they need to select the tab. This is likely only to occur if someone arrives on a page from a google search because if they start from the beginning of the tutorial, they will have had to select scala somewhere. This is why I have a concern about the usability of inline conditionals.

docs-source/docs/modules/microservices-tutorial/pages/complete-entity.adoc
```diff
@@ -302,7 +308,7 @@ Scala::
 include::example$03-shopping-cart-service-scala/src/main/scala/shopping/cart/ShoppingCartServiceImpl.scala[tag=checkoutAndGet]
 ----
 
-ifdef::todo[TODO: refer to the source download, especially for the optional commands that are not shown here]
+As mentioned at the beginning of this page, `RemoveItem` and `AdjustItemQuantity` commands are not mandatory for the subsequent steps of the tutorial. A zip file with the completed for this page implementing the optional commands is available for [.group-scala]#link:_attachments/3-shopping-cart-event-sourced-complete-scala.zip[download]# [.group-java]#link:_attachments/3-shopping-cart-event-sourced-complete-java.zip[download]#.
```
patriknw, line 311: good

docs-source/docs/modules/microservices-tutorial/examples/02-shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCart.java
```diff
@@ -7,16 +7,16 @@
...
 import akka.persistence.typed.PersistenceId;
 import akka.persistence.typed.javadsl.*;
 import com.fasterxml.jackson.annotation.JsonCreator;
 
 import java.time.Duration;
-import java.time.Instant;
-import java.util.*;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
```
patriknw, line 19: Good to not use `*` import. Also in 03 and maybe more places.

patriknw (commented):
I looked at the split command handler builders in 03 ShoppingCart. I agree that it's easier to explain that way (and easier for us when Scala and Java are more similar). `CommandHandlerWithReplyBuilderByState` is quite a mouthful though, but ok.

Let's continue with the split builders.

docs-source/docs/modules/microservices-tutorial/examples/02-shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCart.java
octonato: Yes, I'm planning to do a second pass and organise all the imports. I sent a PR with a formatter (#339) that also touch the imports. However, I didn't check if it as `.*`. I any case, the plugin seems quite configurable. We can probably fix this with the formatter.

docs-source/docs/modules/microservices-tutorial/pages/complete-entity.adoc
octonato, line 311:
```suggestion
As mentioned at the beginning of this page, `RemoveItem` and `AdjustItemQuantity` commands are not mandatory for the subsequent steps of the tutorial. A zip file with the completed steps on this page and implementing the optional commands is available for [.group-scala]#link:_attachments/3-shopping-cart-event-sourced-complete-scala.zip[download]# [.group-java]#link:_attachments/3-shopping-cart-event-sourced-complete-java.zip[download]#.
```

patriknw (approved):
LGTM

Issue #332, opened 2020-10-20
Link checker in PR validation?


patriknw:
Two PRs complaining about links to www.lagomframework.com, which are valid.

Shall we only run it as a cron job?

PR #337, opened 2020-10-20
Run Link Checker only in scheduled builds

ennru:
References #332


ennru:
Cancelled the build manually. The link-checker job wasn't listed.

PR #343, opened 2020-10-21
structure of concepts, #179

patriknw:
Taking some ideas from https://github.com/akka/akka-platform-guide/pull/184 and suggesting this structure (unfolded):

Screenshot 2020-10-21 at 12 11 13

References #179

### DIFF
--- a/docs-source/docs/antora.yml
+++ b/docs-source/docs/antora.yml
@@ -4,5 +4,5 @@ title: "" # removed title to prevent a double title entry in TOC
 nav:
 - modules/ROOT/nav.adoc
 - modules/microservices-tutorial/nav.adoc
-- modules/concepts/nav.adoc
 - modules/how-to/nav.adoc
+- modules/concepts/nav.adoc
--- a/docs-source/docs/modules/concepts/nav.adoc
+++ b/docs-source/docs/modules/concepts/nav.adoc
@@ -1,12 +1,4 @@
-.xref:index.adoc[Concepts]
-* Properties of Reactive Microsystems
-** xref:isolation.adoc[]
-** xref:autonomy.adoc[]
-** xref:mobility.adoc[]
-** xref:message-driven-event-driven.adoc[Event-Driven]
-** xref:state-ownership.adoc[]
-
-* Implementation Concepts
+* Akka modules for Microservices
 ** xref:akka-actor.adoc[]
 ** xref:akka-stream.adoc[]
 ** xref:akka-grpc.adoc[]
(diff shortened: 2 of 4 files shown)


patriknw:
wdyt @rstento @ignasi35 ?

ignasi35:
Looks great. It's very in sync with the contents required for the tutorial and the order in which they appear on the tutorial.

ignasi35 (approved):

octonato:
can we make the last section's title what shorter? 

"Patterns and techniques for Reactive Microservices" is a mouthful. 

I know this was already discussed, but I don't think we came to a conclusion.

Some suggestions: 

* Reactive Microservices Architecture
* Reactive Microservices Concepts
* Reactive Microservices Patterns

I like architecture. What we are listing is a bunch of patterns and techniques that when applied define how your architecture looks like. You can pick a few from the list and if you follow the reactive principles (and Akka helps you on that), you end up with a Reactive Microservices Architecture. 

Maybe the only one that is not fitting here is the deployment. Debatable. 

WDYT?

patriknw:
> can we make the last section's title what shorter?

@ignasi35 had some opinions about that, not sure what it was

Architecture might sound "bigger than it is", and it also has many different meanings for different people

Reactive Microservices Patterns could work, but I think Ignasi thought some of these are not patterns. For example the last about Docker/Kubernetes doesn't fully fit in here. Maybe we should leave that one out (from nav) now, and introduce when we know more about how we will do deployments?

octonato:
> Reactive Microservices Patterns could work, but I think Ignasi thought some of these are not patterns. For example the last about Docker/Kubernetes doesn't fully fit in here. Maybe we should leave that one out (from nav) now, and introduce when we know more about how we will do deployments?

if that's the reason for not choosing "Patterns", then yes let's leave it out. It probably deserve its own section. 

Could "Reactive Microservices Concepts" be a better choice? Pattern is also overloaded. Concepts is generic enough.

ignasi35:
We discussed a few other formulas in https://github.com/akka/akka-platform-guide/pull/184#discussion_r486421919 and https://github.com/akka/akka-platform-guide/pull/184#discussion_r486832089. Here are a few of those formulas:

* `How to design Reactive` <-- can be confused with the `How to...` section of the guide
* `Designing Reactive`
* `Useful for Reactive`

Or a new one I think we didn't consider yet:

* `Designing Reactive Microservices` 

So the NAV would look like:

* Akka modules for Microservices
* What is Reactive?
* Designing Reactive Microservices

octonato:
Designing Reactive Microservices is also good to me

patriknw:
+1 to Designing Reactive Microservices

rstento:
+1 to Designing Reactive Microservices

rstento (approved):
LGTM when "Designing Reactive Microservices" replaces the longer title.

octonato (approved):

PR #345, opened 2020-10-21
use tabset for Maven/sbt in dev-env.adoc and template.adoc, #343

patriknw:
* because there hasn't been any source code choice yet it could be easy to miss
  the language choice
* leaving later sbt/mvn snippets as is, but when/if we add gradle we will probably
  use tabs in all places

References #343

If trying this you should use `make local-preview` instead of opening the files in browser.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
@@ -1,10 +1,14 @@
 = I: Set up your development environment
 :page-supergroup-java-scala: Language
+:page-supergroup-maven-sbt: Build tool
 
 include::ROOT:partial$include.adoc[]
 
+Select your preferred language (Java/Scala) above.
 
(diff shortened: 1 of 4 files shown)


docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -1,11 +1,15 @@
 = I: Set up your development environment
 :page-supergroup-java-scala: Language
+:page-supergroup-maven-sbt: Build tool
```
octonato, line 3: Two selectors will make it confusing, IMO. 

Choosing `scala` should imply `sbt` and choosing `java` implies `mvn`. 

That's the combination we have in the templates. I think the page should stick with it.

patriknw: Then the tab titles can’t be Maven/sbt but ok I’ll change

octonato: You mean they need to be Java/Scala? 

Then we will have a problem when we introduce gradle. 

Maybe we should have no tabs for the build tool and change the language selector to a better position. See https://github.com/lightbend/antora-ui-lightbend-theme/issues/73

patriknw: That's what I said in the ticket, from the beginning. We can use Java/Scala for now, and I think the tabs serve the purpose on these initial pages to make it more evident that there is a choice.

octonato (approved):

PR #357, opened 2020-10-22
add .gitignore to all samples

octonato:
The original template had a .gitignore file that so how didn't propagated to the templates we created here. 

Minor thing, but quite useful when going through all the tutorial steps.


patriknw (approved):
LGTM

PR #361, opened 2020-10-23
more edits from reviewing tutorial

patriknw:

PR #366, opened 2020-10-23
more edits from code review

patriknw:
* final projection-grpc-client.adoc page
* also incorporated feedback from PR https://github.com/akka/akka-platform-guide/pull/361 by @ignasi35: eba456c
* and change the syntax for the source file names in the tabs as suggested by @ignasi35:  c627bf4
  (the reason I didn't use that syntax was that I thought we had tried that and it blended into the source code, but it works fine)

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
@@ -16,6 +16,7 @@ import akka.stream.alpakka.cassandra.javadsl.CassandraSessionRegistry;
 // tag::SendOrderProjection[]
 import shopping.order.proto.ShoppingOrderService;
 import shopping.order.proto.ShoppingOrderServiceClient;
+
 // end::SendOrderProjection[]
 
 // tag::ItemPopularityProjection[]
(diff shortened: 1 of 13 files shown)


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
```diff
@@ -64,11 +65,13 @@ public Main(ActorContext<Void> context) {
...
     ShoppingOrderService orderService = orderServiceClient(system);
     SendOrderProjection.init(system, orderService);
+
     // end::SendOrderProjection[]
   }
 
   // can be overridden in tests
   // tag::SendOrderProjection[]
+  // can be overridden in tests
   protected ShoppingOrderService orderServiceClient(ActorSystem<?> system) {
```
ignasi35, line 75: This looks duplicate :-)

ignasi35 (approved):
Silly nitpick. Otherwise LGTM

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
patriknw: ah, I wanted it to be shown in snippet. I'll fix in next PR

PR #367, opened 2020-10-23
update README files

patriknw:
The Scala projects were mostly up to date. Java projects now have corresponding but with `mvn`


ignasi35 (approved):
LGTM

PR #374, opened 2020-10-23
removeTags is the problem

patriknw:


patriknw:
merging since we will only see if this fixes it from the main build

patriknw:
that worked, created issue https://github.com/akka/akka-platform-guide/issues/375

Issue #352, opened 2020-10-22
failed: IntegrationTest - Address already in use


patriknw:
03 Shopping cart service (Java)

```
Running shopping.cart.IntegrationTest
290[2020-10-22 09:43:44,243] [INFO] [akka.event.slf4j.Slf4jLogger] [] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Slf4jLogger started
291[2020-10-22 09:43:45,673] [INFO] [akka.remote.artery.tcp.ArteryTcpTransport] [akka://IntegrationTest] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Remoting started with transport [Artery tcp]; listening on address [akka://IntegrationTest@127.0.0.1:44325] with UID [-6022436335997463111]
292[2020-10-22 09:43:45,756] [INFO] [akka.cluster.Cluster] [akka://IntegrationTest@127.0.0.1:44325] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Cluster Node [akka://IntegrationTest@127.0.0.1:44325] - Starting up, Akka version [2.6.10] ...
293[2020-10-22 09:43:46,131] [INFO] [akka.cluster.Cluster] [akka://IntegrationTest@127.0.0.1:44325] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Cluster Node [akka://IntegrationTest@127.0.0.1:44325] - Registered cluster JMX MBean [akka:type=Cluster,port=44325]
294[2020-10-22 09:43:46,131] [INFO] [akka.cluster.Cluster] [akka://IntegrationTest@127.0.0.1:44325] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Cluster Node [akka://IntegrationTest@127.0.0.1:44325] - Started up successfully
295[2020-10-22 09:43:46,398] [INFO] [akka.cluster.Cluster] [akka://IntegrationTest@127.0.0.1:44325] [] [IntegrationTest-akka.actor.default-dispatcher-5] - Cluster Node [akka://IntegrationTest@127.0.0.1:44325] - No seed-nodes configured, manual cluster join required, see https://doc.akka.io/docs/akka/current/typed/cluster.html#joining
[...]

PR #376, opened 2020-10-26
Only remove tags for downloadable code

### LINKED ISSUE #375: prepare-downloads.sh should only modify copy of files
removeTags removes on the sources that are used by the docs and then tags are missing resulting in missing snippets

ignasi35:
Fixes #375

PS: I haven't tested on macOS as the `find` and `sed` syntax is slightly different (and it'd fail anyway).

### DIFF
--- a/scripts/prepare-downloads.sh
+++ b/scripts/prepare-downloads.sh
@@ -7,10 +7,9 @@ declare -r temporal_folder="${PWD}/target/zips"
 
 mkdir -p ${tutorial_attachments}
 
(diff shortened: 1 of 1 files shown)


patriknw (approved):
LGTM

patriknw:
Unrelated test failure https://github.com/akka/akka-platform-guide/issues/352

patriknw:
Build was ok now. I'll merge and we can verify the published result.

Issue #382, opened 2020-10-27
Styling of callouts


ennru:
Improve the contrast for the call-outs (eg "(2)"), they hide in the source code. In earlier versions they were inversed.

![image](https://user-images.githubusercontent.com/458526/97309754-b952b900-1862-11eb-90d8-ec4a890024fc.png)

cc @rasummer

ennru:
Nice!

PR #384, opened 2020-10-27
Latest CSS and Playbook changes

rasummer:
@ennru 

### Latest CSS

Updated CSS styles for callouts #382 so that they standout more.

screenshot-localhost_8000-2020 10 27-15_12_33

### Updated YML Playbook Config

Rather than point to the latest Antora default ui zip, which will change as versions are released. We now point to a ui_bundle.zip that lives in the [Theme development repo](https://github.com/lightbend/antora-supplemental-ui-lightbend-theme) at the following public URL:

`https://lightbend.github.io/antora-supplemental-ui-lightbend-theme/default_ui/ui-bundle.zip`

When Antora upgrades it's default UI theme to a new version, we will first test it in the Theme repo against our supplemental customizations.


ignasi35 (approved):

ignasi35:
> We now point to a ui_bundle.zip that lives in the Theme development repo at the following public URL:

When is that `.zip` used? It's downloaded and expanded during the build of the site, right? It'd be great if it eventually becomes a versioned URL so builds are repeatable but we can fix that in the future.

rasummer:
> When is that `.zip` used? It's downloaded and expanded during the build of the site, right? 

Correct

> It'd be great if it eventually becomes a versioned URL so builds are repeatable but we can fix that in the future.

Agreed, I have looked for a versioned zip from Antora, but couldn't find it. I plan to ping them about it.

PR #386, opened 2020-10-29
Styling: update css and bundle zip

rasummer:
@ignasi35 @ennru @rstento 

I'm now versioning the `theme.min.css` files so that we can more easily track what changes in each update. Also easier to rollback for you guys. I'm storing the versions here:  https://github.com/lightbend/antora-supplemental-ui-lightbend-theme/tree/master/dist/css . I will add notes to each new version that describes what changed or was fixed. 

Also I'm now pointing to a `ui-bundle.zip` that has an artifact id in it. Antora's Default UI doesn't currently have versions/releases of the zip, but I can download artifacts of their latest builds named `bundle-stable` - as found here: https://gitlab.com/antora/antora-ui-default/-/jobs. 

I then store them here: https://github.com/lightbend/antora-supplemental-ui-lightbend-theme/tree/master/default_ui and reference Antora's build in the readme. This gives us a little more insight to the "version/age" of the default UI that we are using, easier to roll back etc. 

When we want to upgrade to a newer version of their default-ui, I will create a new artifact version, host it and test against the Lightbend Theme before coming over and updating it in here.


rstento (approved):
LGTM!

ennru (approved):
LGTM.

Issue #303, opened 2020-10-15
"Edit This Page" is a link of type file://


ignasi35:
See it in https://developer.lightbend.com/docs/akka-platform-guide/microservices-tutorial/dev-env.html

![Screenshot 2020-10-15 at 16 40 57](https://user-images.githubusercontent.com/762126/96145337-634a5100-0f05-11eb-9bf4-abe70609bfab.png)

ennru:
What Antora configuration is needed? @rasummer / @rstento

rstento:
@ennru, you set the value in the playbook, the files we have named site.yml and author-mode-site.yml. Here is the doc on the attribute to set,  edit_url: https://docs.antora.org/antora/2.3/playbook/content-edit-url/

My understanding is that you set this for the root directory of the source files in the git repo and Antora will adjust the link to point at the source file of the page you are looking at.

ennru:
The URL gets rendered correctly if I set `url` to the git repository.
When using the local files with `url: ./../` Antora renders the file URL as the docs state. But I can't get the `edit_url` setting to have any effect. (eg. `edit_url: ~` should switch it off altogether)

Setting `CI=true` doesn't have an effect either.

ennru:
@rstento I wasn't able to make this work. Do you have more pointers?

rstento:
@ennru, I also had trouble getting the edit_url: ~ to shut the link off in the past. Maybe we should file an issue? What behavior do you want? Do you want to shut it off altogether, because we can also do that with template.

ennru:
We'd like to have the links to the source files.

rstento:
Is the repo private? I saw that that that can cause a problem.

ennru:
No, this is public. (For a private repo those links are not as useful, indeed.)

rstento:
@ennru, so I had the same problem in the Akka Serverless doc and I fixed it. There we wanted the Edit this page link to only show for the "dev" version which we actually publish at a different URL, but we don't want it to show on prod because it is a commercial product and we don't want anyone contributing to content. I did this:

in my author-mode-site.yml (named dev.yml in our repo) I added the git url:
content:
  sources:
  - url: https://github.com/lightbend/cloudstate-docs.git
    branches: [HEAD]
in the site.yml (named prod.yml in our repo, I was then able to disable the Edit this page link with this:

content:
  edit_url: ~
  sources:
  - url: https://github.com/lightbend/cloudstate-docs.git
    branches: [master]

[...]

PR #387, opened 2020-11-02
Show the real site to Gustav (including edit links)

ennru:
I still don't get the "Edit this page" links to work with the relative sources url, but it works with the real git url using the main branch.
This makes the preview deploy to Gustav use the `site.yml` so no TODOs will show anymore.

References #303


patriknw (approved):
LGTM

PR #389, opened 2020-11-03
Review of the last two pages and making head #2s consistent in form.

rstento:
References #xxxx

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
@@ -33,7 +33,7 @@ Scala::
 ****
 
 :sectnums:
-== Add the gRPC service
+== Add the Order gRPC service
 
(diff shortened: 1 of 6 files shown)


docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
```diff
@@ -33,36 +32,36 @@ Scala::
 * link:_attachments/6-shopping-cart-complete-scala.zip[Source] with the steps on this page completed.
 ****
 
+:sectnums:
+== Add the gRPC service
```
patriknw, line 36: Since we already have one gRPC service, the shopping cart service, it can be good to clarify that this is another.
Could be "Add order gRPC service"?

patriknw (commented):
excellent, just something small...

docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
rstento: Good idea, done.

patriknw (approved):
LGTM, thanks again

PR #396, opened 2020-11-10
change naming convention for Bootstrap service-name

patriknw:
Corresponding to what the operator will use.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-order-service-scala/src/main/resources/local-shared.conf
+++ b/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-order-service-scala/src/main/resources/local-shared.conf
@@ -5,7 +5,7 @@ akka.management.http.hostname = "127.0.0.1"
 ## The following setting overrides for Akka Management, Akka Discovery and Akka Projections
 ## should only be used in a local environment during development.
 akka.management.cluster.bootstrap.contact-point-discovery {
-  service-name = "shoppingorderservice"
+  service-name = "shopping-order-service"
   discovery-method = config
   # boostrap filters ports with the same IP assuming they are previous instances running on the same node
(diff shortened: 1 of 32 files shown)


docs-source/docs/modules/microservices-tutorial/examples/00-shopping-order-service-scala/src/main/resources/local-shared.conf
```diff
@@ -18,8 +18,7 @@ akka.management.cluster.bootstrap.contact-point-discovery {
 }
 
 akka.discovery.config.services {
-  // service name == lower case ActorSystem name by default
-  "shoppingorderservice" {
+  "shopping-order-service" {
```
ignasi35, line 21: Maybe we can keep the comment with the recommended convention?

ignasi35:
```suggestion
   // (recommended) service name == kebab case ActorSystem name
   "shopping-order-service" {
```

ignasi35 (approved):
LGTM

docs-source/docs/modules/microservices-tutorial/examples/00-shopping-order-service-scala/src/main/resources/local-shared.conf
patriknw: When we specify the explicit `service-name` it has no connection with the ActorSystem name anymore, so I don't think we need to confuse the user with any additional recommendation. When we generate the template the user will only specify "shopping-order-service" and we will generate the ActorSystem name with those conventions.

patriknw:
I'll look into the test failures, could be some real mistake

ignasi35 (approved):

PR #398, opened 2020-11-19
Prepare downloads from make

ennru:
To create the source download bundles for TechHub, they need to be packaged via `make`.

We do not need to publish to Gustav anymore. Those URLs redirect to TechHub.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+version: ~> 1.0
+
 language: scala
 services:
   - docker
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -172,11 +172,10 @@ jobs:
...
 
-    - stage: publish
-      script:
-        - scripts/prepare-downloads.sh
-        - scripts/deploy-site.sh
-      name: "Publish snapshot site"
+#    - stage: publish
+#      script:
+#        - scripts/deploy-site.sh
+#      name: "Publish snapshot site"
```
patriknw, line 178: if these are kept as commented we should have a comment why

johanandren: Or delete if we know that they won't be needed. They'll still be in git history so not lost.

johanandren (approved):
LGTM with the commented snippet clarified or removed.

octonato (approved):

patriknw (approved):
LGTM

PR #406, opened 2020-11-20
java/scala link to full download

patriknw:
* link was only to scala zip
* since this is at the very beginning it's easy to miss the java/scala choice so placed intabset as in other download places


ennru (approved):
LGTM.

PR #414, opened 2020-11-30
Link validator: fail on broken links

ennru:
* Upgrade the Link Validator
* Run it on PRs
* Remove Asciidoc link checker

### DIFF
--- a/.github/workflows/link-validator.yml
+++ b/.github/workflows/link-validator.yml
@@ -1,13 +1,10 @@
 name: Link Validator
 
 on:
-  # TODO remove `on push` and `on pull_request_target`
-  push:
-    branches: [main]
   pull_request_target:
     branches: [main]
   schedule:
-    - cron:  '0 6 * * *'
+    - cron:  '0 6 * * 1-5'
 
 jobs:
   validate-links:
@@ -15,6 +12,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Run Antora
         run: make all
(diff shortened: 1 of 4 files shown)


.github/workflows/link-validator.yml
```diff
@@ -1,20 +1,18 @@
...
-    - cron:  '0 6 * * *'
+    - cron:  '0 6 * * 1-5'
 
 jobs:
   validate-links:
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        ref: ${{ github.event.pull_request.head.sha }}
```
octonato, line 15: For what is that good for? 

Is that related or you just spot it was missing and is taking the opportunity to fix it in this PR.

octonato (approved):
LGTM

Left a question though. Not blocking as I think it's unrelated to the PR.

.github/workflows/link-validator.yml
patriknw: The defaults fo `actions/checkout@v2` is stupid for PR validation. Without this it will use the commit of the `main` branch.

.github/workflows/link-validator.yml
```diff
@@ -1,20 +1,18 @@
...
 
 on:
-  # TODO remove `on push` and `on pull_request_target`
-  push:
-    branches: [main]
   pull_request_target:
     branches: [main]
   schedule:
-    - cron:  '0 6 * * *'
+    - cron:  '0 6 * * 1-5'
```
patriknw, line 7: how does cron and pull_request_target relate? For cron jobs I thought it would not be for main and not PRs?

ennru: It will trigger on both, pull requests that target main and on schedule.
The first to ensure PRs don't introduce broken links, the schedule to discover links that go away.
Let's see how this works out.

.github/workflows/link-validator.yml
ennru: Added a link to the relevant issue in the checkout action.

patriknw (approved):
LGTM

PR #416, opened 2020-12-01
sync dependencies

patriknw:
The updates in https://github.com/akka/akka-platform-guide/pull/405 were not at all places.

* update http, grpc, junit
* scala.binary.version in more pom.xml


octonato (approved):

patriknw:
Don't merge yet. In another project with those versions I got
```
You are using version 10.2.1 of Akka HTTP, but it appears you (perhaps indirectly) also depend on older versions of related artifacts. You can solve this by adding an explicit dependency on version 10.2.1 of the [akka-parsing, akka-http2-support, akka-http-jackson] artifacts to your project. See also: https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html#mixed-versioning-is-not-allowed
	at akka.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:176)
	at akka.util.ManifestInfo.checkSameVersion(ManifestInfo.scala:154)
	at akka.http.scaladsl.HttpExt.<init>(Http.scala:79)
```

patriknw:
It was a stupid mistake by me. Forgot clean when building the docker image.

PR #419, opened 2020-12-01
Adding note about breaking test.

jtownley:
Adding note about other tests that break as part of the above changes.

References #xxxx


patriknw (approved):
LGTM, thanks

PR #421, opened 2020-12-04
Use Akka and Akka HTTP BOM pom.xml

ennru:
This is WIP until Akka and Akka HTTP release their next versions.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-analytics-service-java/pom.xml
+++ b/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-analytics-service-java/pom.xml
@@ -56,11 +56,6 @@
             <url>https://repo.akka.io/snapshots/</url>
             <layout>default</layout>
         </repository>
-        <repository>
-            <id>akka-http-snapshots</id>
-            <name>Akka HTTP Snapshots</name>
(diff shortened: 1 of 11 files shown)


patriknw (commented):
nice!

ignasi35 (approved):
I think this can be simplified further thanks to the power the BOM brings, actually.

For example, the explicit dependency to `akka-http-jackson_{scala.bin}` was necessary to fix an incompatibility with a transitive dependency brought in by akka-management. Maybe `akka-discovery_` can be gone too.
I suspect we may also be able to remove `akka-http2-support_{scala.bin}` (and other on the `akka-http` set) since I suspect that explicit dependency is only necessary to fix the issue when bringing in akka-grpc.

ignasi35 (changes requested):
(pressed the wrong button, sorry)

Apparently, when the PR is DRAFT I can't dismiss my own Approval, or maybe that's normal limitation?

ennru:
@ignasi35 You are right, all explicit dependencies to Akka HTTP can go as Akka gRPC brings them in and the BOM makes sure we get the correct version.

ignasi35 (approved):

ennru:
See #466 instead

PR #422, opened 2020-12-04
JDBC: Spring Data JPA integration

patriknw:
Food for thought. I'm not decided myself if this is a good idea or if plain JPA/Hibernate is better.

* I also tried reactive repositories, but that's not for Spring Data JPA
* I also tried the `@Async` annotation but that resulted in weird runtime exceptions (class cast exc)


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularityProjection.java
```diff
@@ -42,10 +56,12 @@ public static void init(ActorSystem<?> system, ItemPopularityRepository reposito
             CassandraReadJournal.Identifier(), // <4>
             tag);
 
-    return CassandraProjection.atLeastOnce( // <5>
+    return JdbcProjection.exactlyOnce( // <5>
         ProjectionId.of("ItemPopularityProjection", tag),
         sourceProvider,
-        () -> new ItemPopularityProjectionHandler(tag, repository)); // <6>
+        () -> new HibernateJdbcSession(entityManagerFactory.createEntityManager()),
```
patriknw, line 62: I haven't verified if the transactions (and EntityManager instance) is propagated correctly from the Projection to the repository yet. I guess something more is needed, unless it's magic.

octonato:
I did some changes here, main points are:

* added create-tables.sql
* Uses `persistence.xml` to configure db. That's how the java community does it, so let's follow their habits
* write side is using the jdbc plugin
* one must start the docker-compose to get postgres running

TODO
- [ ] refactor tests to use JDBC

**Road block** 

The JDBC plugin doesn't have an option to create the schema and that is important for testing (see https://github.com/akka/akka-persistence-jdbc/issues/69). 

Moreover, the Cassandra plugin has the keyspace and each test runs on a different keyspace. For JDBC, we need a different DB each time (eg: `jdbc:h2:mem:test1`, `jdbc:h2:mem:test2`, etc.) and we need to be able to tell it to the Spring layer. Surely possible, but we need some more infra before we get there.


ps: this PR is now point to new `jdbc-example` branch.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<persistence xmlns="http://java.sun.com/xml/ns/persistence"
+             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+             xsi:schemaLocation="http://java.sun.com/xml/ns/persistence http://java.sun.com/xml/ns/persistence/persistence_2_0.xsd"
+             version="2.0">
+    <persistence-unit name="shopping-cart-hibernate">
+        <properties>
+            <property name="javax.persistence.jdbc.driver" value="org.postgresql.Driver"/>
+            <property name="javax.persistence.jdbc.url" value="jdbc:postgresql://localhost:5432/shopping-cart?reWriteBatchedInserts=true"/>
```
patriknw, line 9: Can these be replaced with System.properties, environment variables or something? The operator will provide this the url and the credentials.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/persistence.conf
```diff
@@ -1,45 +1,48 @@
...
 }
 
-# tag::healthchecks-for-cassandra[]
-akka.management {
-  health-checks {
-    readiness-checks {
-      akka-persistence-cassandra = "akka.persistence.cassandra.healthcheck.CassandraHealthCheck"
+akka-persistence-jdbc {
+  shared-databases {
+    shopping-cart {
```
patriknw, line 20: would be good to use a generic name here because the operator will replace url, user and password.
See https://github.com/lightbend/akka-platform-operator/pull/229/files#diff-782583be98d2215e3c4e1e5a6c40e6dcf20a2731cf9631574948e7f3164ef1eaR16
and https://github.com/lightbend/akka-platform-operator/pull/228/files#diff-9e40cc95712f299023366811c8d994eb6eecf7c98b46b8f2181a3b468bc577f5R215

we could use the CR name `shopping-cart-name` if it's important to have something more unique, but probably easiest with generic name

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/META-INF/persistence.xml
octonato: I'm having trouble with this setup for tests and was already considering going back to properties. Didn't think on System.properties though, but that's a good point. We should be able to replace them all.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/persistence.conf
octonato: sure, we can could follow Play's convention. The DB is always called `default`, unless explicitly changed by user.

patriknw: maybe nice if we read them from the ActorSystem.settings.config so that they are only defined at one place, i.e. back to the programatic spring setup for these properties

octonato: ok, got the Config wired inside Spring. 

next step: get the test working.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/ddl-scripts/create_tables.cql
```diff
@@ -64,19 +64,19 @@ CREATE TABLE IF NOT EXISTS metadata(
...
+      timestamp bigint,
+      ser_id int,
+      ser_manifest text,
+      snapshot_data blob,
+      snapshot blob,
+      meta_ser_id int,
+      meta_ser_manifest text,
+      meta blob,
+      PRIMARY KEY (persistence_id, sequence_nr))
+      WITH CLUSTERING ORDER BY (sequence_nr DESC);
```
patriknw, line 79: accidental change? revert

(we should remove this script from the example, it should be in the cassandra example)

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/ddl-scripts/create_tables.sql
```diff
@@ -0,0 +1,35 @@
...
+    "CURRENT_OFFSET" VARCHAR(255) NOT NULL,
+    "MANIFEST" VARCHAR(4) NOT NULL,
+    "MERGEABLE" BOOLEAN NOT NULL,
+    "LAST_UPDATED" BIGINT NOT NULL
+    );
+
+create index "PROJECTION_NAME_INDEX" on public."AKKA_PROJECTION_OFFSET_STORE" ("PROJECTION_NAME");
+
+alter table public."AKKA_PROJECTION_OFFSET_STORE"
+    add constraint "PK_PROJECTION_ID" primary key("PROJECTION_NAME","PROJECTION_KEY");
```
patriknw, line 35: there should also be a create_user_tables.sql with the item_popularity table

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml
```diff
@@ -19,16 +19,27 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <scala.binary.version>2.13</scala.binary.version>
         <akka.version>2.6.10</akka.version>
-        <akka-http.version>10.2.1</akka-http.version>
+        <akka-http.version>10.2.0</akka-http.version>
```
patriknw, line 22: revert

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml
```diff
@@ -19,16 +19,27 @@
...
         <akka-management.version>1.0.9</akka-management.version>
-        <akka-persistence-cassandra.version>1.0.4</akka-persistence-cassandra.version>
         <alpakka-kafka.version>2.0.5</alpakka-kafka.version>
         <akka-projection.version>1.0.0</akka-projection.version>
-        <akka-grpc.version>1.0.2</akka-grpc.version>
+        <akka-grpc.version>1.0.1</akka-grpc.version>
```
patriknw, line 26: revert

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/AsyncItemPopularityRepository.java
```diff
@@ -0,0 +1,21 @@
+package shopping.cart;
```
patriknw, line 1: what do you think about having a new package `shopping.cart.repository`?

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularityRepository.java
```diff
@@ -1,11 +1,13 @@
...
 import java.util.Optional;
-import java.util.concurrent.CompletionStage;
 
-public interface ItemPopularityRepository {
-  CompletionStage<Done> update(String itemId, int delta);
+public interface ItemPopularityRepository extends Repository<ItemPopularity, String> {
+
+  ItemPopularity save(ItemPopularity itemPopularity);
+
+  Optional<ItemPopularity> findById(String id);
```
patriknw, line 11: nice, ah so it's by naming conventions when extending `Repository`, and no additional annotations are needed

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/AsyncItemPopularityRepository.java
patriknw: and we can place the `SpringIntegration` and `HibernateJdbcSession` there also

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/SpringConfig.java
```diff
@@ -0,0 +1,86 @@
...
+    factory.setDataSource(dataSource());
+    factory.setJpaProperties(additionalProperties());
+
+    return factory;
+  }
+
+
+  @Bean
+  public DataSource dataSource() {
+    // FIXME: this needs to be wrapped in a connection pool
```
patriknw, line 52: ok, then we should look into that. (Could be a separate PR). I assume there is something in Spring for it.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/SpringConfig.java
```diff
@@ -0,0 +1,86 @@
...
+    dataSource.setUsername(config.getString("database.user"));
+    dataSource.setPassword(config.getString("database.password"));
+    return dataSource;
+  }
+
+
+  Properties additionalProperties() {
+    Properties properties = new Properties();
+
+    Config additionalProperties = this.config.getConfig("database.additional-properties");
```
patriknw, line 65: good

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/local-shared.conf
```diff
@@ -43,5 +43,8 @@ shopping-order-service.host = "localhost"
 shopping-order-service.port = 8301
 // end::grpc[]
 
-# for reduced Projection latency
-akka.persistence.cassandra.events-by-tag.eventual-consistency-delay = 200 ms
+database {
+  url = "jdbc:postgresql://localhost:5432/shopping-cart?reWriteBatchedInserts=true"
+  user = "shopping-cart"
+  password = "shopping-cart"
+}
```
patriknw, line 50: add newline

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/local-shared.conf
```diff
@@ -43,5 +43,8 @@ shopping-order-service.host = "localhost"
 shopping-order-service.port = 8301
 // end::grpc[]
 
-# for reduced Projection latency
-akka.persistence.cassandra.events-by-tag.eventual-consistency-delay = 200 ms
+database {
```
patriknw, line 46: could we rename this to `jdbc-connection-settings`, so that it matches with other things (kafka-connection-settings) in the Operator

https://github.com/lightbend/akka-platform-operator/pull/228/files#diff-9e40cc95712f299023366811c8d994eb6eecf7c98b46b8f2181a3b468bc577f5R209

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/resources/persistence.conf
```diff
@@ -1,45 +1,73 @@
...
-
-      query {
-        refresh-interval = 2s
-      }
-
-      journal.keyspace = "shoppingcartservice"
-      snapshot.keyspace = "shoppingcartservice"
-    }
+    journal.plugin = "jdbc-journal"
+    snapshot-store.plugin = "jdbc-snapshot-store"
```
patriknw, line 5: auto-start also
```
    journal {
       plugin = "jdbc-journal"
       auto-start-journals = ["jdbc-journal"]
     }
     snapshot-store {
       plugin = "jdbc-snapshot-store"
       auto-start-snapshot-stores = ["jdbc-snapshot-store"]
     }
```

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/test/java/shopping/cart/IntegrationTest.java
```diff
@@ -46,28 +50,11 @@
...
-                + KEYSPACE
-                + "\n"
-                + "akka.projection.cassandra.offset-store.keyspace = "
-                + KEYSPACE
-                + "\n"
-                + "shopping-cart-service.kafka.topic = \"shopping_cart_events_"
-                + UNIQUE_QUALIFIER
-                + "\"")
-        .withFallback(ConfigFactory.load("integration-test.conf"));
+    return ConfigFactory.parseResources("integration-test.conf").resolve();
```
patriknw, line 54: why not `ConfigFactory.load` ?

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityProjectionTest.java
```diff
@@ -1,52 +1,71 @@
...
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import org.junit.ClassRule;
-import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
 
 public class ItemPopularityProjectionTest {
   // stub out the db layer and simulate recording item count updates
   static class TestItemPopularityRepository implements ItemPopularityRepository {
```
patriknw, line 29: good

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/test/resources/persistence-test.conf
```diff
@@ -0,0 +1,5 @@
+database {
+  url = "jdbc:postgresql://localhost:5433/shopping-cart-test?reWriteBatchedInserts=true"
+  user = "shopping-cart"
+  password = "shopping-cart"
+}
```
patriknw, line 5: line break

patriknw (commented):
looking very good, only some small things...

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/test/java/shopping/cart/IntegrationTest.java
octonato: Should also work with `.load`. I think I needed to change to parse because at some point and need to combine a few of them with variable substitution and then resolve. Probably not needed anymore.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml
octonato: any reason for reverting it?

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/ddl-scripts/create_tables.sql
octonato: The table is created by hibernate. We can added the script, but we need to keep the auto-create for tests. Can be done by setting the properties in tests.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/ddl-scripts/create_tables.cql
octonato: yeah, remove better

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml
patriknw: Why would we not use latest version? this is changing from 1.0.2 to 1.0.1.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/ddl-scripts/create_tables.sql
patriknw: yes, that should only be for tests
would be strange if we have some tables auto created and some not

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/pom.xml
octonato: Right, I read it the other way around. 
I don't know how this got in.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularityRepository.java
octonato: `Repository` is a marker, but there is somewhere a list of methods that the different implementations of `Repository` can implement. `CrudRepository` implements a few. 

If you stick with the convention, the foreseen impl is picked dynamically for you. 

Crazy, no? 🤪

octonato:
@patriknw, I did another round here and applied all the suggestions. 

The tests are are broken when run all together. They pass if run individually and with a clean DB (new dockers). The reason is that we don't drop the offset store table at the beginning of the test. We do it for the journal (`dropAndCreate`), but not for projections. 

So, a test can see the offset-store from the previous test and if they happen to share the same projeciton_name, the second test won't consume the events it is expected to consume. 
I'm talking here about integration tests. So the projection name is fixed and is shared and we can't tweak it (not desirable).

I think the solution here is to include a `dropIfExists` method in the projection testkit. Btw, maybe `dropAndCreate` should go and we should use `dropIfExists` and `createIfNotExists` everywhere. It also better communicate what happens behind the scene. 

`dropIfExists` will also solve the issue I discover today with MySQL and index creation.

patriknw:
I fixed the `CreateTableTestUtils` and verified that tests are ok, and local run.

patriknw (commented):
LGTM

PR #425, opened 2020-12-15
Deploy WIP branch to akka.io (#424)

ennru:
Include #424 in `main-with-operator`.

PR #431, opened 2020-12-17
Debug deploy WIP branch to akka.io 

ennru:
For some reason base64 reports `base64: invalid input` when decoding the secret.

https://github.com/akka/akka-platform-guide/runs/1569028576?check_suite_focus=true#step:3:156

References #425

### DIFF
--- a/scripts/deploy-site.sh
+++ b/scripts/deploy-site.sh
@@ -5,6 +5,7 @@ if [ -z ${SCP_SECRET} ]; then
 fi
 make build-wip
 eval "$(ssh-agent -s)"
-echo ${SCP_SECRET} | base64 -d > /tmp/id_rsa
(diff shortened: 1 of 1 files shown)


scripts/deploy-site.sh
```diff
@@ -5,6 +5,7 @@ if [ -z ${SCP_SECRET} ]; then
 fi
 make build-wip
 eval "$(ssh-agent -s)"
-echo ${SCP_SECRET} | base64 -d > /tmp/id_rsa
+printf "%.10s\n" ${SCP_SECRET}
```
patriknw, line 8: will this reveal the real content of the secret?

ennru: Yes, the first 10 characters.

PR #438, opened 2020-12-18
Java tutorial steps using JDBC integration

octonato:
References #427

Issue #429, opened 2020-12-16
How-to use Cassandra instead of RDBMS


patriknw:
The tutorial example will use RDBMS (jdbc). We should have a how-to describing how to use Cassandra as an alternative.

The full example zip should be available with Cassandra but we don't need to keep it with Cassandra for each step.

PR #440, opened 2020-12-18
Replace Cassandra by Postgres

octonato:
Draft because on top of #438.

This is still very WIP. I only replaced some basic references to Cassandra by Postgres. 

The most challenging will the `projection-query.adoc`. This page introduces the first projection and it should cover all the Spring Data integration for Java and then whatever we pick for Scala. 

It will take use some iterations to find a good way to build that page. 

As an addition challenge, I tried to use some [.group-java] to add some paragraphs only for Java the layout broke. Probably some mix that 

I didn't keep the Cassandra pages as discussed. I don't think it makes much sense. We said we want to add an 'how-to' appendix and I don't think we can re-use much of what we have done for Cassandra. The references to Cassandra are spread all over the tutorial pages. We need to start a new page showing the  refactorings needed to use Cassandra. 

References #427


docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
```diff
@@ -7,7 +7,7 @@ Select your preferred language (Java/Scala) above.
...
 
-This tutorial uses Cassandra and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker {tab-icon}, window="tab"]. The template project's root directory includes a https://docs.docker.com/compose/[Docker Compose {tab-icon}, window="tab"] file that you can use to run Cassandra and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
+This tutorial uses a Postgres database and Kafka. You may run local installations during development, but we recommend running both services using https://docs.docker.com[Docker {tab-icon}, window="tab"]. The template project's root directory includes a https://docs.docker.com/compose/[Docker Compose {tab-icon}, window="tab"] file that you can use to run Postgres and Kafka. The provided Docker Compose file also runs Zookeeper which is required to run a Kafka broker.
```
patriknw, line 10: The big question is if we can make the [Scala RDBMS example](https://github.com/akka/akka-platform-guide/issues/429) in this sprint or not? If we can't make it then we have to keep the Cassandra for Scala and use group-scala/group-java conditions.

docs-source/docs/modules/microservices-tutorial/pages/projection-query.adoc
```diff
@@ -45,7 +45,7 @@ To process events in a projection, we will:
 
 * implement the event processing of the Projection in a `Handler`
 * encapsulate database access with `ItemPopularityRepository`, which can have a stubbed implementation for tests
-* add an implementation for Cassandra
+* add an implementation for JDBC [.group-java]#using Spring Data and Hibernate#
```
patriknw, line 48: I think we can include the Spring "infrastructure" code in the project template. Then we don't have to explain it in detail (with full copy paste snippets) in this page. Later we might move some of that code to a akka-projection-spring-data module.

docs-source/docs/modules/microservices-tutorial/pages/dev-env.adoc
patriknw: As we discussed last Friday we will take a stab at the Scala RDBMS example so this PR can assume that we use Postgres

octonato:
This is good for a first round. 

I managed to move things around enough that we can accomodate a Scala JDBC variant without too much difference in files.

docs-source/docs/modules/microservices-tutorial/pages/projection-query.adoc
```diff
@@ -258,23 +286,30 @@ sbt compile
...
 +
+.src/main/java/shopping/cart/ShoppingCartServiceImpl.java
 [source,java,indent=0]
 ----
 include::example$04-shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCartServiceImpl.java[tag=getItemPopularity]
 ----
++
+<1> add the `ItemPopularityRepository` to the service implementation constructor
+<2> use the `ActorSystem` to get an instance of an `Executor` tailored for blocking operations
+<3> Implement `getItemPopularity` by calling the repository to find the projected model by id. Note that calls to repository are blocking therefore we wrap the call using `CompletableFuture.supplyAsync` and we use the blocking executor to run it.
```
patriknw, line 303: We should highlight this more than in a callout. A warning, which also explains why this is important (thread starvation of other tasks). That can be a follow up in separate PR.

patriknw (approved):
LGTM

docs-source/docs/modules/microservices-tutorial/pages/projection-query.adoc
octonato: Sure. I tried to add something like that but had issues with formatting in Antora. I will try harder and maybe add the call out in another place (not inside the tabset.

patriknw: yes, it shouldn't be inside the tabset. we will need similar for Scala

patriknw:
merge?

octonato:
> merge?

No, we need a green build.

patriknw:
ok, strange that the scala projects have jdbc stuff already?

octonato:
> ok, strange that the scala projects have jdbc stuff already?

My bad. Sleep deprived, doing crazy this week. 🤪

PR #444, opened 2020-12-26
Update projection-grpc-client.adoc

sgc109:


patriknw (approved):
LGM, thanks for contributing

PR #451, opened 2021-01-11
upgrade to ccpa cookie banner

rasummer:
@ennru @rstento 

This updates the theme to the latest cookie banner, which uses geo-location to present Californian users with a CCPA specific preference center and cookie banner. It also upgrades the theme to the latest CSS and JavaScript, which is need for the banner upgrade to function.


ennru (approved):
Thanks.

PR #455, opened 2021-01-13
Fix persistence config loading

octonato:
`application.conf` can't have unresolved values because Slick is eagerly loading it. 

We need to first load the `local-share.conf` and then the `application.conf` so in dev-mode user gets the connection settings defined correctly.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/src/main/resources/local1.conf
+++ b/docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/src/main/resources/local1.conf
@@ -1,8 +1,7 @@
-include "application"
 include "local-shared"
+include "application"
 
 shopping-cart-service.grpc.port = 8101
 
 akka.remote.artery.canonical.port = 2551
 akka.management.http.port = 9101
-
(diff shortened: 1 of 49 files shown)


docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/src/main/resources/local1.conf
```diff
@@ -1,8 +1,7 @@
-include "application"
 include "local-shared"
+include "application"
```
patriknw, line 2: The intention was that the values in local-shared.conf should override the values in application.conf.
How does that work now?

ennru: The first definition wins in Config.

ennru (approved):
LGTM.

docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/src/main/resources/local1.conf
patriknw: that is not correct, see https://github.com/akka/akka-platform-guide/pull/456

PR #464, opened 2021-01-15
Fix Cassandra health-check include

octonato:
I removed too much on my previous PRs. 

Once we move to Scala JDBC, we will need to keep it around somewhere until we get back the full Cassandra example.


patriknw (approved):
LGTM

PR #465, opened 2021-01-15
Minor improvements on Entity page

octonato:

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
@@ -71,6 +71,7 @@ include::example$02-shopping-cart-service-scala/src/main/scala/shopping/cart/Sho
 [.tabset]
 Java::
 +
+.src/main/java/shopping/cart/ShoppingCart.java
 [source,java,indent=0]
 ----
 include::example$02-shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCart.java[tags=shoppingCart;events]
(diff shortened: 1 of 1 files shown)


docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
```diff
@@ -164,24 +175,28 @@ To glue the command handler, event handler, and state together, we need some ini
...
+.src/main/java/shopping/cart/ShoppingCart.java
 [source,java,indent=0]
 ----
 include::example$02-shopping-cart-service-java/src/main/java/shopping/cart/ShoppingCart.java[tags=shoppingCart;init]
 ----
++
+<1> The entities are distributed over the nodes in the Akka Cluster with Cluster Sharding.
+<2> An `EventSourcedBehavior` is created for the `ShoppingCart`.
+<3> Snapshotting is an optimization to reduce recovery when the entity is started.
+<4> Restarting with backoff in case of failures.
```
octonato, line 187: It's better to show the callouts inside the tabset. The drawback is that we need to repeat it for java and scala. 

The advantage is that they become better associated with the code.

Here we see it inside the border.
CleanShot 2021-01-15 at 14 20 59

Here is when added outside the tabset. 
CleanShot 2021-01-15 at 14 21 47


Moreover, some snippets do need different callouts, so better to align the style in all pages. Even if sometimes we need to repeat text.

patriknw: changing that at all places seems to be a cosmetic change that can wait

patriknw (approved):
LGTM

PR #467, opened 2021-01-15
show how to configure kubectl for eks

octonato:


patriknw (approved):
LGTM, I have never had to do that, but good to include

octonato:
> LGTM, I have never had to do that, but good to include

But you need to point kubectl to the cluster somehow. You probably have done it at some point.  
There are other ways of doing it, but that's what I found on AWS documentation.

PR #469, opened 2021-01-18
How-to use Cassandra instead of RDBMS, #429

patriknw:
How-to guide of using Cassandra instead of PostgreSQL.

The example src is pretty much identical copy from `main` so you don't have to review that.

References #429


docs-source/docs/modules/how-to/pages/cassandra-alternative.adoc
```diff
@@ -0,0 +1,502 @@
+= How to use Cassandra instead of PostgreSQL
```
patriknw, line 1: This is the doc page that should be reviewed.

octonato (approved):
LGTM

PR #477, opened 2021-01-19
update dependencies

patriknw:


ignasi35 (approved):
LGTM

Side note: it may be time for an akka-projection release. Plus, there are three different akka-projection versions used across the repo (1.0.0, 1.0.0+27 and 1.0.0+33)

octonato:
> Side note: it may be time for an akka-projection release. Plus, there are three different akka-projection versions used across the repo (1.0.0, 1.0.0+27 and 1.0.0+33)

Absolutely, I was planning to do it last week but side-tracked.

PR #482, opened 2021-01-19
minor doc adjustments

patriknw:

PR #474, opened 2021-01-18
Adds "Run in Kubernetes" section for JDBC

octonato:
References #449 

- [x] Add Run in Kubernetes for Projection to Kafka
- [x] Add Run in Kubernetes for Projection to other gRPC service

PR #486, opened 2021-01-20
Place anchors above titles

octonato:
Moves all anchors to above the title as discussed here: https://github.com/akka/akka-platform-guide/pull/482#discussion_r560443378

I didn't change the grpc service because PR #474 is already doing that.


patriknw (approved):
LGTM

PR #489, opened 2021-01-20
IAM for metering, only MeterUsage required

chbatey:


patriknw (approved):
LGTM, have you updated the global AkkaPlatformOperator policy so that we are testing with this?

chbatey:
Yep

PR #494, opened 2021-01-21
clarify RDS and MSK instructions

patriknw:
Clarified the RDS security group settings to be more similar to the MSK instructions.

Still something weird with the RDS security groups that need to be sorted out.


patriknw:
More will come in separate PR

PR #496, opened 2021-01-21
simplify RDS instructions

patriknw:
* VPC / security group instructions didn't work for external access
* use psql via EKS Pod instead of pgAdmin to avoid exernal access (would be difficult to configure)
* same approach with kubectl run as for MSK


octonato (approved):
👍

chbatey (approved):

PR #498, opened 2021-01-21
improve Run in Kubernetes instructions

patriknw:
* verified all steps, including all grpcurl commands

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
+++ b/docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
@@ -361,7 +361,6 @@ mvn compile exec:exec -DAPP_CONFIG=local1.conf
 
 === Exercise the service
 
-// # tag::exercise[]
 Use `https://github.com/fullstorydev/grpcurl[grpcurl]` to exercise the service:
 
 . Use `grpcurl` to add 3 socks to a cart:
@@ -388,7 +387,6 @@ grpcurl -d '{"cartId":"cart1", "itemId":"t-shirt", "quantity":2}' -plaintext 127
 ----
 +
 The returned updated cart should still contain the 3 socks.
-// # end::exercise[]
 
 === Exercise with multiple service instances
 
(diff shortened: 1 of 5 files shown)


docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
```diff
@@ -493,7 +491,30 @@ kubectl get akkamicroservices/shopping-cart-service
 
 include::partial$prepare-to-exercise-in-kube.adoc[]
 
-include::entity.adoc[tag=exercise]
+Use `https://github.com/fullstorydev/grpcurl[grpcurl]` to exercise the service:
```
patriknw, line 494: can't use same, local has ctrl-c instructions

docs-source/docs/modules/microservices-tutorial/pages/projection-kafka.adoc
```diff
@@ -318,9 +318,23 @@ Create a xref:deployment:aws-install.adoc[Kubernetes cluster and install the Akk
...
+[source,shell script]
+----
+kubectl create secret generic shopping-cart-service-kafka-secret --from-literal=bootstrapServers=<copied bootstrap servers connect string>
+----
+
+and another secret (with same content) for the `shopping-analytics-service`:
+
+[source,shell script]
+----
+kubectl create secret generic shopping-analytics-service-kafka-secret --from-literal=bootstrapServers=<copied bootstrap servers connect string>
```
patriknw, line 334: easy to miss that two secrets must be created

docs-source/docs/modules/microservices-tutorial/pages/projection-grpc-client.adoc
```diff
@@ -354,19 +352,32 @@ kubectl logs -f <shopping-cart-service pod name from above>
...
 kubectl port-forward svc/shopping-cart-service-grpc 8101:8101
 ----
+
+Use `https://github.com/fullstorydev/grpcurl[grpcurl]` to exercise the service:
+
+. Use the checkout in the shopping cart service with `grpcurl`:
 +
 [source,shell script]
 ----
-kubectl port-forward svc/shopping-order-service-grpc 8301:8301
```
patriknw, line 366: the port forward to order-service is just adding complexity without value

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
octonato: I was trying to avoid this kind of repetition and make sure that both exercises were the same. Not a big issue though, we are talking about a few repeated lines. 

An alternative would be to not include ctrl-c instructions and then we say: "stop the service, restart it and verify that data has been persisted".

Then, just after the "exercise" block we add a hint:

for local:

> use ctrl-c to stop the service and restart it

for remote: 

> remove the CR to bring the service down and redeploy it

As it is now, we are not testing / demonstrating the persistent part in K8s. Again, I don't think is a big deal.

octonato (approved):
LGTM

chbatey (approved):

docs-source/docs/modules/microservices-tutorial/pages/entity.adoc
patriknw: It's kind of tested since the response includes previous item from the cart (unless the implementation would keep it memory). Not a big deal as you say. Don't want to add too many steps.

PR #502, opened 2021-01-22
use new releases for jdbc plugin and projections

octonato:
First commit updates the seeds, second commit is just the copy-files. 

So, better to review only bb9a1cfd12ca06daa798f9cf2cfba04b0700dd0c


patriknw:
I would expect some more in the how-to examples

octonato:
> I would expect some more in the how-to examples

What you mean by some more? We have been using the the snapshots for a while already and using the new schema in the examples. We are only moving now to the stable release.

patriknw:
More dependencies. For example docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/pom.xml

octonato:
> More dependencies. For example docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/pom.xml

yep, I missed those. Will fix it.

patriknw (approved):
LGTM

PR #508, opened 2021-01-22
sync dependencies

patriknw:
* order and analytics still had Casssandra

I used diff with shopping-cart-service and made them as identical as possible


octonato (approved):
LGTM

Issue #109, opened 2020-08-27
failed: IntegrationSpec


patriknw:
In Travis, see attached log 
[IntegrationSpec.log](https://github.com/akka/akka-platform-guide/files/5136075/IntegrationSpec.log)

PR #512, opened 2021-01-25
Use Kustomize to customize the service installation

andreaTP:
Follow up.

### DIFF
--- a/docs-source/docs/modules/deployment/pages/operator-reference.adoc
+++ b/docs-source/docs/modules/deployment/pages/operator-reference.adoc
@@ -62,6 +62,60 @@ For multi-threaded applications such as the JVM, the CFS scheduler limits are an
 
 To avoid CFS scheduler limits, it is best not to use `resources.limits.cpu` limits, but use `resources.requests.cpu` configuration instead.
 
+== Configure logback
+
+A custom logback configuration can be provided to the deployment, make sure that your application is already configured to use the {akka}/typed/logging.html#slf4j-backend[Slf4j backend {tab-icon}, window="tab"] for logging.
+
+Create a logback configuration you want to apply in a `logback.xml` file:
+
+[source,xml]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>EASY LOGGING - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="TRACE">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+</configuration>
+----
+
+Create the secret in the cluster:
+
+[source,shell script]
+----
+kubectl create secret generic my-logback --from-file=logback.xml=./logback.xml
+----
+
+Modify the deployment descriptor to use this configuration:
+
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  logbackSecret:
+    secretName: "my-logback"
+----
+
+
+Apply the logback configuration by running:
+
+[source,shell script]
+----
+kubectl apply -f kubernetes/shopping-cart-service-cr.yml
+----
+
+And the operator is going to take care of applying it.
+
 == Namespaces
 
 The Akka Operator manages Akka Microservices in all namespaces of the Kubernetes cluster, which means that you should only install the operator once. It is possible to configure the operator to only manage a specific namespace if that would be needed.
--- a/docs-source/docs/modules/how-to/nav.adoc
+++ b/docs-source/docs/modules/how-to/nav.adoc
@@ -4,3 +4,4 @@
 * xref:scale-independently.adoc[scale command and query-side]
 * xref:enable-TLS.adoc[enable TLS for gRPC]
 * xref:health-checks.adoc[inspect the health]
+* xref:configure-deployments.adoc[customize AkkaMicroservice deployment]
--- a/docs-source/docs/modules/how-to/pages/configure-deployments.adoc
+++ b/docs-source/docs/modules/how-to/pages/configure-deployments.adoc
@@ -1,0 +1,82 @@
+= Configure Akka Microservices deployments
+
+
+include::partial$include.adoc[]
+
+== Customize AkkaMicroservices deployments
+
+To apply customizations on top of a default deployment descriptor (CR) we recommend using https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/[Kustomize], built into `kubectl`.
+
+Prepare a folder called `base` containing the following files:
+
+The default deployment descriptor (CR):
+
+.base/shopping-cart-service.yml
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  image: <image>
+----
+
+A basic reference to it for Kustomize:
+
+.base/kustomization.yml
+[source,yaml,title='']
+----
+resources:
+- shopping-cart-service.yml
+----
+
+You can already deploy your default service by running:
+
+[source,shell script]
+----
+kubectl apply -k base
+----
+
+== Apply customizations
+
+If you want to deploy, e.g., to an environment with limited resources you can tweak just a few parameters in the CR.
+
+Create a separate folder along `base` and call it `custom` with this content:
+
+The actual customization:
+
+.custom/custom.yml
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  replicas: 1
+  resources:
+    limits:
+      memory: 512Mi
+    requests:
+      cpu: 500m
+      memory: 512Mi
+----
+
+And a reference to the customization:
+
+.custom/kustomization.yml
+[source,yaml]
+----
+bases:
+- ../base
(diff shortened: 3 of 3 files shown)


docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -11,28 +11,69 @@ The https://lightbend.github.io/akka-operator-helm/akka-operator/crds/v1/akka-mi
 
 A fictive example of a https://lightbend.github.io/akka-operator-helm/api-reference/v1/shopping-cart-service-full-cr.yml[deployment descriptor (CR) containing all properties {tab-icon}, window="tab"].
 
+To apply customizations on top of a default deployment descriptor (CR) we suggest to leverage https://kustomize.io[Kustomize] built into `kubectl`.
```
seglo, line 14:
```suggestion
To apply customizations on top of a default deployment descriptor (CR) we recommend using https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/[Kustomize], built into `kubectl`.
```

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -11,28 +11,69 @@ The https://lightbend.github.io/akka-operator-helm/akka-operator/crds/v1/akka-mi
...
+----
+
+A basic reference to it for Kustomize:
+[source,yaml,title='kustomization.yml']
+----
+resources:
+- shopping-cart-service.yml
+----
+
+You can already deploy you default service by running:
```
seglo, line 35:
```suggestion
Deploy your default service by running:
```

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -11,28 +11,69 @@ The https://lightbend.github.io/akka-operator-helm/akka-operator/crds/v1/akka-mi
...
+And a reference to the customization:
+[source,yaml,title='kustomization.yml']
+----
+bases:
+- ../base
+patchesStrategicMerge:
+- replicas.yaml
+----
+
+You can scale the number of pods by applying the changed deployment descriptor by running:
```
seglo, line 68:
```suggestion
Scale the number of pods by applying the overlaid deployment descriptor by running:
```

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -11,28 +11,69 @@ The https://lightbend.github.io/akka-operator-helm/akka-operator/crds/v1/akka-mi
...
 
 [source,shell script]
 ----
-kubectl apply -f kubernetes/shopping-cart-service-cr.yml
+kubectl apply -k replicas
 ----
 
+Kubernetes will add or remove Pods to adjust to the desired number.
```
seglo, line 75:
```suggestion
Kubernetes will reconcile the number of Pods to the desired scale.
```

seglo (approved):
LGTM. A few suggestions.

docs-source/docs/modules/deployment/pages/operator-reference.adoc
patriknw, line 14: I suggest that we describe the kustomize approach in a separate section and keep the examples here as they were. I see kustomize as an optional approach.

It can be a separate `==` section in this `operator-reference.adoc`. At some point I think we should extract it into a [how-to guide instead](https://akka.io/platform-guide/wip/how-to/index.html) but it's ok to keep all these additional topics in operator-reference.adoc for now.

andreaTP: My personal bias is that the kustomize approach is arguably cleaner, more "production-ready" and helps the user to follow the most recent kubernetes best practices.
Sticking with `kubectl apply -f` we decide to favor retro-compatibility and a more bare-bone "demo" approach.

But having to keep and maintain the 2 different approaches seems like unnecessary work and maintenance.
In case we prefer `kubectl apply -f` I'm ok with dropping this PR.

RayRoestenburg: +1 on kustomize approach

patriknw: I want the examples to be as simple as possible, introducing as few new tools as possible. We can revisit that later but I want us to keep it simple to begin with.

However, I think it's great to describe the possibility.

RayRoestenburg: It's important to note that kustomize is not another tool. It is a feature of kubectl that is just there.

andreaTP: > I want the examples to be as simple as possible

Having less code copy-pasted makes them easier and nicer to follow IMO.

> introducing as few new tools as possible

kustomize is embedded in kubectl, we are not introducing new tools.
We are, as well, following the lead of kubernetes official documentation e.g.:
https://kubectl.docs.kubernetes.io/guides/config_management/secrets_configmaps/#configmapsgenerator
https://kubectl.docs.kubernetes.io/guides/app_deployment/publishing_bases/#publishing-a-white-box-base

andreaTP: @patriknw for context I have opened #515 based on this approach.
Having to describe the process without kustomize is going to be much more awkward and involved.

patriknw: > kustomize is embedded in kubectl, we are not introducing new tools.

I know, but in my opinion it's adding additional complexity if we would use it in all tutorial examples. More difficult to understand how things are merged together than just one yml file.

andreaTP: We are going in loops, I do understand where you come from @patriknw but I do not share the same opinion.

In this case, I read `additional complexity` as `separation of concerns` that is actually desirable.

I have the impression that @RayRoestenburg and @seglo share my same opinion, but, let's agree to disagree and move on.
If we don't take this approach I need more guidance on how you would structure and phrase i.e. #515 to be able to follow up (i.e. do you want to include instructions on how to base64 the secret? ...).
Thanks!

patriknw: what do you mean with base64? secret can be created with
```
kubectl create secret generic shopping-cart-service-astra-secret --from-literal=username=scott --from-literal=password=tiger --from-file=secureConnectBundle=/path/to/secure-connect.zip
```

patriknw: > In this case, I read additional complexity as separation of concerns that is actually desirable.

I'm talking about the tutorial example: https://akka.io/platform-guide/wip/microservices-tutorial/grpc-service.html#kubernetes

The additional kuztomize file and mandatory directory structure would introduce complexity to that example, without immediate value for those that are learning this for the first time.

chbatey: I agree that kustomize makes it harder for users new to Kubernetes and those users that haven't adopted it; requiring a user to understand resources and that they can be composed from multiple files is more to grasp than here's everything in one file, go and apply it to your cluster.

I don't want us to have to explain kustomize in this guide or for a user to have to go and learn what it is. Best practice defined by blogs/conferences doesn't mean the majority of k8s users use it yet. The separation of concerns makes it simpler, but not easier. 

That said the point of the operator, IMO, is to make deploying Akka easier with fewer things required in resources as we have sensible defaults / manage many of resources for you. So I don't think it matters which approach we show in the guide so we should decide quickly and move on.

RayRoestenburg: Kustomize provides an easy to learn way to organize the resource you want to apply, including secrets that you refer to. 
No everything is defined in one file, not even right now.

> Best practice defined by blogs/conferences doesn't mean the majority of k8s users use it yet.
Hm.... An assumption that can go either way. We're not advocating for it because we read it on a blog obviously, but because it is an easy way to structure more than one resource in a convenient way, and many have used this on the job. Many kubectl invocations, for the user to copy paste is not very useful.

RayRoestenburg: I would say use kustomize in the examples where it makes sense. There is no point in kustomize if there is nothing to customize, very obviously. But where it is useful we should not be scared of it. The user is already supposedly fine with many kubectl commands. They will copy / paste whatever we give them right?

RayRoestenburg: Have we thought of supporting distinct environments? dev / stage / prod / etc. Users will likely need that once they stop playing around, just copy/pasting commands from our docs. This can be nicely structured with kustomize, so getting the user accustomed to that is not a bad thing?

patriknw: > I would say use kustomize in the examples where it makes sense. There is no point in kustomize if there is nothing to customize, very obviously. But where it is useful we should not be scared of it.

That's what I have been trying to say.  I suggested that we describe the kustomize usage in a separate how-to section (or separate section in this operator-reference.adoc). Leaving the examples in the tutorial as plain cr-yml files and kubectl commands for creating secrets.

> Have we thought of supporting distinct environments? dev / stage / prod / etc. Users will likely need that once they stop playing around, just copy/pasting commands from our docs.

That can also be part of that kustomize how-to. Later, I'm pretty sure we will have stories around our recommendations for CI/CD and then that might play an important role (or not). We are not at that working on that story yet (well the story might not even exist yet).

andreaTP:
moved to a separate "how to" section ...
Couldn't test the result since Antora keeps breaking ...

cc. @patriknw @RayRoestenburg

patriknw (approved):
LGTM, thanks

patriknw:
The test failure is unrelated #109

PR #515, opened 2021-01-26
Instructions on how to configure logback

andreaTP:
Opening this PR to show some of the advantages of using the "kustomize" approach in the documentation.
In this case:
 - the secret content is in clear text
 - no need to use bash tricks to generate the secret (e.g. `base64` etc.)
 - exclusively plain kubectl commands
 - no separate workflow for the CR and the secret

### DIFF
--- a/docs-source/docs/modules/deployment/pages/operator-reference.adoc
+++ b/docs-source/docs/modules/deployment/pages/operator-reference.adoc
@@ -62,6 +62,60 @@ For multi-threaded applications such as the JVM, the CFS scheduler limits are an
 
 To avoid CFS scheduler limits, it is best not to use `resources.limits.cpu` limits, but use `resources.requests.cpu` configuration instead.
 
+== Configure logback
+
+A custom logback configuration can be provided to the deployment, make sure that your application is already configured to use the {akka}/typed/logging.html#slf4j-backend[Slf4j backend {tab-icon}, window="tab"] for logging.
+
+Create a logback configuration you want to apply in a `logback.xml` file:
+
+[source,xml]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>EASY LOGGING - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="TRACE">
+        <appender-ref ref="STDOUT"/>
+    </root>
+
+</configuration>
+----
+
+Create the secret in the cluster:
+
+[source,shell script]
+----
+kubectl create secret generic my-logback --from-file=logback.xml=./logback.xml
+----
+
+Modify the deployment descriptor to use this configuration:
+
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  logbackSecret:
+    secretName: "my-logback"
+----
+
+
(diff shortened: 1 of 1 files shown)


RayRoestenburg (approved):
LGTM

RayRoestenburg:
I really like the approach, very clear!

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -123,6 +123,66 @@ For multi-threaded applications such as the JVM, the CFS scheduler limits are an
...
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  logbackSecret:
+    secretName: "my-logback"
+----
+
+the reference to the customization:
+[source,yaml,title='kustomization.yml']
```
patriknw, line 144: I don't know what `title` does, but the way we have it in other places would be:
```
.kubernetes/kustomization.yml:
[source,yaml]
```

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -123,6 +123,66 @@ For multi-threaded applications such as the JVM, the CFS scheduler limits are an
...
+----
+bases:
+- ../base
```
patriknw, line 147: Does it have to use this base directory structure?
I would like that the examples fit in with the examples from the tutorial, which are not using kustomize.
Could it apply the overlay on top on the ordinary `kubernetes/shopping-cart-service-cr.yml` ?

docs-source/docs/modules/deployment/pages/operator-reference.adoc
andreaTP: I can switch to this encoding np 👍

docs-source/docs/modules/deployment/pages/operator-reference.adoc
andreaTP: No, this doesn't work AFAIK, you need a `kustomization.yaml` to reference the cr

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -123,6 +123,66 @@ For multi-threaded applications such as the JVM, the CFS scheduler limits are an
 
 To avoid CFS scheduler limits, it is best not to use `resources.limits.cpu` limits, but use `resources.requests.cpu` configuration instead.
 
+== Configure logback
+
+A custom logback configuration can be provided to the deployment, make sure that your application is already configured to https://doc.akka.io/docs/akka/current/logging.html#slf4j[use Slf4j for logging] .
```
chbatey, line 128: That's classic logging, you don't need anything put a lgoback impl added to the classpath now: https://doc.akka.io/docs/akka/current/typed/logging.html#slf4j-backend

chbatey (commented):
I am happy to move to kustomize in the guide but think we should change all of it if we do

docs-source/docs/modules/deployment/pages/operator-reference.adoc
patriknw: and links to Akka should be like:
```
{akka}/typed/persistence.html#effects-and-side-effects[Akka reference documentation {tab-icon}, window="tab"]
```

andreaTP:
Switched to `kubectl apply -f` approach.
cc. @patriknw @RayRoestenburg

RayRoestenburg (approved):

patriknw (approved):
LGTM, thanks

PR #521, opened 2021-01-28
sketch api for Durable State Actors

octonato:


docs-source/docs/modules/microservices-tutorial/examples/shopping-order-service-scala/src/main/scala/akka/persistence/typed/scaladsl/DurableStateBehavior.scala
```diff
@@ -0,0 +1,75 @@
...
+  def persist[State](state: State): StateEffectBuilder[State] = Persist(state)
+
+  /**
+   * Major difference with respect to EventSourced, a CRUD model can be deleted
+   *
+   * A delete must trigger a stop or reload the empty state
+   * In any case, it needs to start from scratch.
+   */
+  def delete[State](): StateEffectBuilder[State] =
+    Delete.asInstanceOf[StateEffectBuilder[State]]
```
octonato, line 51: My main concern with the `delete` case is that if the actor handles a new message, it needs to be on initial state as if it was loaded for the first time. 

After `delete` completes, the entry must have gone from the DB (hard delete). A new instantiation would eventually re-create it. 

If users want to preserve the fact the there was something that was deleted, they will need to perform an update instead and put a tombstone in the place. Anyway, basic CRUD semantics.

The problem for us is that the actor will work as a data case for that row, so we need to invalidate it somehow.

docs-source/docs/modules/microservices-tutorial/examples/shopping-order-service-scala/src/main/scala/shopping/order/ShoppingOrder.scala
```diff
@@ -0,0 +1,72 @@
...
+  trait Command extends CborSerializable
+
+  trait Response extends CborSerializable
+  case class MissingOrder(id: String, msg: String) extends Response
+  case class CurrentOrder(id: String, item: Map[String, Int]) extends Response
+
+  case class CreateOrder(id: String, items: Map[String, Int]) extends Command
+  case class ReadOrder(id: String, replyTo: ActorRef[Response]) extends Command
+  case class UpdateOrder(id: String, items: Map[String, Int]) extends Command
+  case class DeleteOrder(id: String, replyTo: ActorRef[String]) extends Command
```
octonato, line 19: CRUD goodies!

octonato:
I'm closing this PR in favor of https://github.com/akka/akka/pull/29984

Issue #522, opened 2021-01-28
Issues pulling Adopt Open JDK when following instructions


WadeWaldron:
### Versions used 

Akka Platform Guide (WIP)

### Expected Behavior

I was following the documentation here:

https://akka.io/platform-guide/wip/microservices-tutorial/grpc-service.html
https://akka.io/platform-guide/wip/deployment/aws-install.html
https://akka.io/platform-guide/wip/deployment/aws-ecr.html

Following this, I expect to reach a point where I can build my application docker image without any issues.

### Actual Behavior

When I reached the following step, things broke down:

https://akka.io/platform-guide/wip/microservices-tutorial/grpc-service.html#_build_docker_image

The specific error was:

```
[ERROR] DOCKER> Unable to pull 'adoptopenjdk:11-jre-hotspot' from registry '803424716218.dkr.ecr.eu-central-1.amazonaws.com' : {"message":"repository 803424716218.dkr.ecr.eu-central-1.amazonaws.com/adoptopenjdk not found: name unknown: The repository with name 'adoptopenjdk' does not exist in the registry with id '803424716218'"} (Not Found: 404) [{"message":"repository 803424716218.dkr.ecr.eu-central-1.amazonaws.com/adoptopenjdk not found: name unknown: The repository with name 'adoptopenjdk' does not exist in the registry with id '803424716218'"} (Not Found: 404)]
```

I was able to resolve the error by running:

```
docker pull adoptopenjdk:11-jre-hotspot
```

So as long as you have the necessary image pulled locally, there is no problem. But when you don't it seems that things don't work correctly.

### Reproducible Test Case

Delete `adoptopenjdk:11-jre-hotspot` from your local docker images.

Run the following:

```
mvn -DskipTests -Ddocker.registry=803424716218.dkr.ecr.eu-central-1.amazonaws.com clean package docker:push
```

patriknw:
@WadeWaldron it works if the full image name is specified in pom.xml
```
docker.io/library/adoptopenjdk:11-jre-hotspot
```

PR #530, opened 2021-01-29
full name of jre image, #522

patriknw:
* otherwise mvn docker:push doesn't work without docker pull first when logged in to ecr repo

References #522


octonato (approved):
👍

PR #547, opened 2021-02-04
Akka Operator Cluster status docs

ygree:
Docs for Akka management cluster status reporting.

### DIFF
--- a/docs-source/docs/modules/deployment/pages/operator-reference.adoc
+++ b/docs-source/docs/modules/deployment/pages/operator-reference.adoc
@@ -123,13 +123,13 @@ The Akka Operator will automatically mount the `logback.xml` and include system 
 
 == Akka Cluster status
 
-Akka Operator checks and reports the Akka Cluster membership status. It includes the number of replicas that have joined the Akka cluster and their state from the cluster point of view. This can be Up or Non-Up. The total number of replicas is a sum of Up plus Non-Up. Up means that a replica has successfully joined the cluster.
+Akka Operator checks and reports the Akka cluster membership status. It includes the number of members that have joined the Akka cluster and their state from the Akka cluster point of view. This can be Up or Non-Up. The total number of members is a sum of Up plus Non-Up. Up means that a replica has successfully joined the cluster.
 
-Comparing the number of Up nodes to the current number of replicas can provide an understanding of whether all running replicas are in a normal state or not. Similarly, the number of Non-Up replicas tells us how many replicas are not in the normal operating state. 
+Comparing the number of Up nodes to the current number of replicas can provide an understanding of whether all running replicas are in a normal state or not. Similarly, the number of Non-Up replicas tells us how many replicas are not in the normal operating state, such as Joining or Exiting.
 
 Additionally to Up and Non-Up, there is also Unreachable. This is the number of unreachable nodes from the view of one (the oldest node). For example, if the oldest node was partitioned from the rest of a cluster of Up nodes it would have a high number of unreachable, while if you ask the other nodes they would only say that one node, the oldest, is unreachable.
 
-These numbers can be useful to watch the Akka Cluster state during rolling updates to make sure it is in a normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
+These numbers can be useful to watch the Akka cluster state during rolling updates to make sure it is in a normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
 
 == Namespaces
 
--- a/docs-source/docs/modules/deployment/pages/operator-reference.adoc
+++ b/docs-source/docs/modules/deployment/pages/operator-reference.adoc
@@ -123,7 +123,13 @@ The Akka Operator will automatically mount the `logback.xml` and include system 
 
 == Akka Cluster status
 
-Akka Operator will check and report the Akka Cluster status. It includes number of replicas that are Up, Non-Up or Unreachable. Comparing it to the current number of replicas can provide understanding whether all running replicas are in normal state or not. It can be useful during rolling updates to make sure that Akka Cluster is in normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
+Akka Operator checks and reports the Akka Cluster membership status. It includes the number of replicas that have joined the Akka cluster and their state from the cluster point of view. This can be Up or Non-Up. The total number of replicas is a sum of Up plus Non-Up. Up means that a replica has successfully joined the cluster.
+
+Comparing the number of Up nodes to the current number of replicas can provide an understanding of whether all running replicas are in a normal state or not. Similarly, the number of Non-Up replicas tells us how many replicas are not in the normal operating state. 
+
+Additionally to Up and Non-Up, there is also Unreachable. This is the number of unreachable nodes from the view of one (the oldest node). For example, if the oldest node was partitioned from the rest of a cluster of Up nodes it would have a high number of unreachable, while if you ask the other nodes they would only say that one node, the oldest, is unreachable.
+
+These numbers can be useful to watch the Akka Cluster state during rolling updates to make sure it is in a normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
 
 == Namespaces
 
--- a/docs-source/docs/modules/deployment/pages/operator-reference.adoc
+++ b/docs-source/docs/modules/deployment/pages/operator-reference.adoc
@@ -121,6 +121,10 @@ kubectl apply -f kubernetes/shopping-cart-service-cr.yml
 
 The Akka Operator will automatically mount the `logback.xml` and include system property `logback.configurationFile` to the application, which means that it will be loaded instead of the default `logback.xml` that is bundled with the application. You have to restart the Pods for it to take effect, which will happen automatically by a rolling update when you apply changes of the deployment descriptor.
 
+== Akka Cluster status
+
+Akka Operator will check and report the Akka Cluster status. It includes number of replicas that are Up, Non-Up or Unreachable. Comparing it to the current number of replicas can provide understanding whether all running replicas are in normal state or not. It can be useful during rolling updates to make sure that Akka Cluster is in normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
+
 == Namespaces
 
 The Akka Operator manages Akka Microservices in all namespaces of the Kubernetes cluster, which means that you should only install the operator once. It is possible to configure the operator to only manage a specific namespace if that would be needed.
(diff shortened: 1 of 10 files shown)


docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -121,6 +121,10 @@ kubectl apply -f kubernetes/shopping-cart-service-cr.yml
 
 The Akka Operator will automatically mount the `logback.xml` and include system property `logback.configurationFile` to the application, which means that it will be loaded instead of the default `logback.xml` that is bundled with the application. You have to restart the Pods for it to take effect, which will happen automatically by a rolling update when you apply changes of the deployment descriptor.
 
+== Akka Cluster status
+
+Akka Operator will check and report the Akka Cluster status. It includes number of replicas that are Up, Non-Up or Unreachable. Comparing it to the current number of replicas can provide understanding whether all running replicas are in normal state or not. It can be useful during rolling updates to make sure that Akka Cluster is in normal state. See xref:troubleshooting.adoc[troubleshooting deployment status] for some examples.
```
johanandren, line 126: Maybe make this even more clear that it is about membership vs replicas - "... the Akka Cluster membership status. It includes the number of replicas that have joined the Akka cluster and their state from the point of view of the cluster."

Also note that unreachable is more like a flag and not a membership status, a node that is unreachable will also be either the up or non up count so it's rather "Up or Non-Up and how many of them are unreachable."

I wonder if we should even point out that the unreachability it is the view of one (the oldest node) to make that extra clear? (For example if the oldest node was partitioned from the rest of a cluster of UP nodes it would have a high number of unreachable, while if you ask the other nodes they would only say that one node, the oldest, is unreachable)

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
```diff
@@ -11,31 +11,44 @@ The `akkamicroservices` CR has a status with the purpose to give an overview of
...
 kubectl get akkamicroservices shopping-cart-service
-NAME                    PHASE     REPLICAS   DESIRED   AGE
-shopping-cart-service   Running   2          2         7h14m
+NAME                    PHASE     REPLICAS   DESIRED   AGE     UP
+shopping-cart-service   Running   2          2         7h14m   1
 ----
 
+- `REPLICAS` displays how many replicas are currently running.
+- `DESIRED` displays the desired number of replicas of the application.
+- `UP` displays how many replicas are in normal operating state.
```
johanandren, line 30: "How many replicas that have successfully joined the Akka cluster"? (Since they can also be unreachable "normal operating state" is not quite true)

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
```diff
@@ -11,31 +11,44 @@ The `akkamicroservices` CR has a status with the purpose to give an overview of
...
-shopping-cart-service   Running   2          2         7h14m
+NAME                    PHASE     REPLICAS   DESIRED   AGE     UP
+shopping-cart-service   Running   2          2         7h14m   1
 ----
 
+- `REPLICAS` displays how many replicas are currently running.
+- `DESIRED` displays the desired number of replicas of the application.
+- `UP` displays how many replicas are in normal operating state. 
+
+If `REPLICAS` equals `UP` then all the Akka cluster is up and running without any issues on all the allocated pods. If `REPLICAS` is bigger than `UP` then not shown by default `NON-UP` and `UNREACHABLE` will provide details on how many replicas are not in normal operating state or unreachable.
```
johanandren, line 32: So again, all could be up but unreachable (at least for a while before SBR likely kicks in and kills them)

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
```diff
@@ -11,31 +11,44 @@ The `akkamicroservices` CR has a status with the purpose to give an overview of
...
-shopping-order-service       Running   1          1         7h9m    Running 1 pods.   803424716218.dkr.ecr.eu-central-1.amazonaws.com/shopping-order-service:20201218-061851-566473f
+NAME                         PHASE     REPLICAS   DESIRED   AGE   UP    NON-UP   UNREACHABLE  MESSAGE           IMAGE
+shopping-analytics-service   Running   1          1         7h17m 1     0        0            Running 1 pods.   803424716218.dkr.ecr.eu-central-1.amazonaws.com/shopping-analytics-service:20201218-061851-566473f
+shopping-cart-service        Running   2          2         7h15m 2     0        0            Running 2 pods.   803424716218.dkr.ecr.eu-central-1.amazonaws.com/shopping-cart-service:20201218-061851-566473f
+shopping-order-service       Running   1          1         7h9m  1     0        0            Running 1 pods.   803424716218.dkr.ecr.eu-central-1.amazonaws.com/shopping-order-service:20201218-061851-566473f
 ----
 
+In this mode, additionally to `UP` there are also `NON-UP` and `UNREACHABLE` that display how many replicas are non-up or unreachable respectively.
+
+- `NON-UP` displays how many replicas are NOT in normal operating state, e.g. leaving, joining or down.
```
johanandren, line 47: Good showing _what_ non up states are here.

johanandren (commented):
Good, I think we can make it a bit more explicit with details though.

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
ygree: Thanks, I'll rephrase it. Just for my understanding. I think it can never be all because at least the oldest can reach itself. Right?

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
ygree: Will fix it.

docs-source/docs/modules/deployment/pages/operator-reference.adoc
ygree: Thanks for your suggestions. I'll fix it.

johanandren (approved):
LGTM

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -121,6 +121,16 @@ kubectl apply -f kubernetes/shopping-cart-service-cr.yml
 
 The Akka Operator will automatically mount the `logback.xml` and include system property `logback.configurationFile` to the application, which means that it will be loaded instead of the default `logback.xml` that is bundled with the application. You have to restart the Pods for it to take effect, which will happen automatically by a rolling update when you apply changes of the deployment descriptor.
 
+== Akka Cluster status
+
+Akka Operator checks and reports the Akka Cluster membership status. It includes the number of replicas that have joined the Akka cluster and their state from the cluster point of view. This can be Up or Non-Up. The total number of replicas is a sum of Up plus Non-Up. Up means that a replica has successfully joined the cluster.
```
patriknw, line 126: s/number of replicas/number of members/

so member is Akka terminology and replicas is Kubernetes

patriknw: s/from the cluster/from the Akka Cluster/

docs-source/docs/modules/deployment/pages/operator-reference.adoc
```diff
@@ -121,6 +121,16 @@ kubectl apply -f kubernetes/shopping-cart-service-cr.yml
...
+
+Comparing the number of Up nodes to the current number of replicas can provide an understanding of whether all running replicas are in a normal state or not. Similarly, the number of Non-Up replicas tells us how many replicas are not in the normal operating state.
```
patriknw, line 128: , such as Joining or Exiting.

patriknw (commented):
good, just something minor

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
johanandren: That is correct, reachability is only about the other nodes in the cluster, while up/members is about this node plus the other nodes.

patriknw (approved):
LGTM, but let's wait with merging until we have the publish from `current` instead of `main` in place

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
```diff
@@ -48,17 +63,17 @@ The PHASE is `Running` when everything working. It is updated when there is some
...
 
 If the deployment failed, for example if referencing a missing secret:
 
 [source,shell script]
 ----
 kubectl get akkamicroservices shopping-cart-service -o wide
-NAME                    PHASE              REPLICAS   DESIRED   AGE     MESSAGE
-shopping-cart-service   DeploymentFailed   2          2         7h25m   akka.operator.action.ActionException: Action [GetAction] failed: jdbc Secret [shopping-cart-service-jdbc-zecret] doesn't exist
+NAME                    PHASE              REPLICAS   DESIRED   AGE   UP    NON-UP   UNREACHABLE  MESSAGE
+shopping-cart-service   DeploymentFailed   2          2         7h25m 1     0        0            akka.operator.action.ActionException: Action [GetAction] failed: jdbc Secret [shopping-cart-service-jdbc-zecret] doesn't exist
```
ignasi35, line 76: This example seems to contradict the sentence above:

"If `REPLICAS` is higher than `UP` then `NON-UP` will represent the non-zero number of replicas that are NOT in a normal operating state. In other words, the sum of `UP` and `NON-UP` equals to `REPLICAS`."

In this case, the numbers and the message are very weird. There's one node that's up (self-joined) even if there's a failure with a missing secret? That looks strange. If this is a situation that may happen, I think we could extend the docs explaining what this really means.

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
```diff
@@ -11,31 +11,46 @@ The `akkamicroservices` CR has a status with the purpose to give an overview of
...
 kubectl get akkamicroservices shopping-cart-service
-NAME                    PHASE     REPLICAS   DESIRED   AGE
-shopping-cart-service   Running   2          2         7h14m
+NAME                    PHASE     REPLICAS   DESIRED   AGE     UP
+shopping-cart-service   Running   2          2         7h14m   1
 ----
 
+- `REPLICAS` displays how many replicas are currently running
+- `DESIRED` displays the desired number of replicas of the application
+- `UP` displays how many replicas have successfully joined the Akka cluster
```
ignasi35, line 30:
```suggestion
- `REPLICAS` displays how many pods are currently running
- `DESIRED` displays the desired number of replicas of the application
- `UP` displays how many nodes have successfully joined the Akka cluster
```

ignasi35 (approved):
let a couple of quick comments.

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
ygree: That's a typo. I haven't noticed that this example originally had 2 replicas. Thanks, I'll fix it.

docs-source/docs/modules/deployment/pages/troubleshooting.adoc
ygree: I think three different terms can be confusing to a reader. Especially the node word that can be understood differently in K8s context.

Issue #534, opened 2021-01-29
failed: ItemPopularityIntegrationTest: expected:<9> but was:<5>


patriknw:
https://github.com/akka/akka-platform-guide/pull/531/checks?check_run_id=1791316864

```
[2021-01-29 12:00:50,365] [WARN] [org.hibernate.engine.jdbc.spi.SqlExceptionHelper] [] [] [ItemPopularityIntegrationTest-akka.projection.jdbc.blocking-jdbc-dispatcher-34] - SQL Error: 0, SQLState: 23505
122
Error: 1-29 12:00:50,365] [ERROR] [org.hibernate.engine.jdbc.spi.SqlExceptionHelper] [] [] [ItemPopularityIntegrationTest-akka.projection.jdbc.blocking-jdbc-dispatcher-34] - ERROR: duplicate key value violates unique constraint "item_popularity_pkey"
123
  Detail: Key (itemid)=(item2) already exists.
124
Error: 1-29 12:00:50,391] [ERROR] [akka.projection.jdbc.internal.JdbcProjectionImpl$JdbcInternalProjectionState] [akka://ItemPopularityIntegrationTest@127.0.0.1:44739] [] [ItemPopularityIntegrationTest-akka.actor.default-dispatcher-22] - [ItemPopularityProjection-carts-1] Failed to process envelope with offset [Sequence(3)]. Projection will stop as defined by recovery strategy.
125
org.springframework.dao.DataIntegrityViolationException: could not execute statement; SQL [n/a]; constraint [item_popularity_pkey]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement
126
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.convertHibernateAccessException(HibernateJpaDialect.java:276)
127
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:233)
128
	at org.springframework.orm.jpa.JpaTransactionManager.doCommit(JpaTransactionManager.java:566)
129
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:743)
130
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:711)
131
	at shopping.cart.repository.HibernateJdbcSession.commit(HibernateJdbcSession.java:61)
132
[...]

PR #552, opened 2021-02-05
fix pk constraint exc in IntegrationTest

patriknw:
References #534

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityIntegrationTest.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityIntegrationTest.java
@@ -104,6 +104,7 @@ public class ItemPopularityIntegrationTest {
     assertEquals(4L, summary3.items.get(item2).intValue());
 
     probe.awaitAssert(
+        Duration.ofSeconds(10),
         () -> {
           Optional<ItemPopularity> item2Popularity = itemPopularityRepository.findById(item2);
           assertTrue(item2Popularity.isPresent());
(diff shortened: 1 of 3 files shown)


docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityIntegrationTest.java
```diff
@@ -104,6 +104,7 @@ public void consumeCartEventsAndUpdatePopularityCount() throws Exception {
     assertEquals(4L, summary3.items.get(item2).intValue());
 
     probe.awaitAssert(
+        Duration.ofSeconds(10),
```
patriknw, line 107: What I think happens is that the test is adding item2 to two different carts and two different projects process those events at the same time, resulting in pk violation. It's supposed to retry, but I think it need some more time for that.

I will update all places if this works.

docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/resources/logback-test.xml
```diff
@@ -7,7 +7,7 @@
         </encoder>
     </appender>
 
-    <root level="INFO">
+    <root level="DEBUG">
```
patriknw, line 10: will revert but wanted to see more if it still fails

ignasi35:
Partially related to this, I think optimistic locking is not working as we expected. I'll try to fix it in #559.

docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityIntegrationTest.java
ignasi35: I did a bit of extra investigating on https://github.com/akka/akka-platform-guide/pull/559 and I think this PR fixes the flakiness.

ignasi35 (approved):

Issue #274, opened 2020-10-09
failed: IntegrationSpec - 84 did not equal 42


patriknw:
In 05-shopping-cart-service

https://travis-ci.com/github/akka/akka-platform-guide/jobs/397295585
```
[info] - should init and join Cluster (5 seconds, 463 milliseconds)
971[2020-10-09 07:40:07,260] [INFO] [akka.kafka.internal.SingleSourceLogic] [akka://IntegrationSpec@127.0.0.1:46509] [] [IntegrationSpec-akka.actor.default-dispatcher-13] - [685c2] Starting. StageActor Actor[akka://IntegrationSpec/system/Materializers/StreamSupervisor-4/$$n#-719105094]
972[2020-10-09 07:40:07,356] [INFO] [org.apache.kafka.clients.consumer.ConsumerConfig] [] [] [IntegrationSpec-akka.kafka.default-dispatcher-33] - ConsumerConfig values: 
973	allow.auto.create.topics = true
974	auto.commit.interval.ms = 5000
975	auto.offset.reset = earliest
976	bootstrap.servers = [localhost:9092]
977	check.crcs = true
978	client.dns.lookup = default
979	client.id = 
980	client.rack = 
981	connections.max.idle.ms = 540000
982	default.api.timeout.ms = 60000
983	enable.auto.commit = false
984	exclude.internal.topics = true
985	fetch.max.bytes = 52428800
986	fetch.max.wait.ms = 500
987	fetch.min.bytes = 1
988	group.id = f7655419-1556-421c-a812-5177dad7e8fc
989	group.instance.id = null
990	heartbeat.interval.ms = 3000
991	interceptor.classes = []
992	internal.leave.group.on.close = true
993	isolation.level = read_uncommitted
994	key.deserializer = class org.apache.kafka.common.serialization.StringDeserializer
995	max.partition.fetch.bytes = 1048576
996	max.poll.interval.ms = 300000
997	max.poll.records = 500
998	metadata.max.age.ms = 300000
999	metric.reporters = []
1000	metrics.num.samples = 2
1001	metrics.recording.level = INFO
1002	metrics.sample.window.ms = 30000
1003	partition.assignment.strategy = [class org.apache.kafka.clients.consumer.RangeAssignor]
1004	receive.buffer.bytes = 65536
1005	reconnect.backoff.max.ms = 1000
1006	reconnect.backoff.ms = 50
1007	request.timeout.ms = 30000
1008	retry.backoff.ms = 100
[...]

PR #555, opened 2021-02-08
all 3 service discovery endpoints in IntegrationTest, #274

patriknw:
* it formed 2 separate clusters, hence the double count

References #274


patriknw:
this is also failing rather frequently, so pls review and merge

ignasi35 (approved):
Oh, wow!

Issue #460, opened 2021-01-14
Align Akka HTTP module versions


johanandren:
Currently there are explicit dependencies on some modules, but not akka-http-core which comes in transitively via akka-management and akka-grpc-runtime which means that if one changes the AkkaHttpVersion it can lead to module version mismatch.

This could also be true for Akka modules but I did not investigate further.

patriknw:
Thanks, let's fix that

PR #563, opened 2021-02-09
Pin down transitive Akka modules

johanandren:
References #460 

I've pinned down _all_ Akka modules even if transitive dependencies on `akka-remote` isn't very likely. I'm thinking better safe than sorry with a sample for beginners. Actually noted quite a few core modules where different other deps are trying to pull the version in one direction or the other (`akka-management`, `akka-projections` and `akka-persistence-jdbc`)

Would be good with someones eye on it before I port it to the maven/java sample and apply to all steps of the guide.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
+++ b/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
@@ -42,25 +42,37 @@ ThisBuild / dynverSeparator := "-"
 
 libraryDependencies ++= Seq(
   // 1. Basic dependencies for a clustered application
+  "com.typesafe.akka" %% "akka-actor-typed" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-actor" % AkkaVersion,
   "com.typesafe.akka" %% "akka-stream" % AkkaVersion,
   "com.typesafe.akka" %% "akka-cluster-typed" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-remote" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-distributed-data" % AkkaVersion,
   "com.typesafe.akka" %% "akka-cluster-sharding-typed" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster-sharding" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster-tools" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-protobuf-v3_2.13" % AkkaVersion,
   "com.typesafe.akka" %% "akka-actor-testkit-typed" % AkkaVersion % Test,
   "com.typesafe.akka" %% "akka-stream-testkit" % AkkaVersion % Test,
   // Akka Management powers Health Checks and Akka Cluster Bootstrapping
   "com.lightbend.akka.management" %% "akka-management" % AkkaManagementVersion,
   "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-http-core" % AkkaHttpVersion,
   "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-parsing" % AkkaHttpVersion,
   "com.lightbend.akka.management" %% "akka-management-cluster-http" % AkkaManagementVersion,
   "com.lightbend.akka.management" %% "akka-management-cluster-bootstrap" % AkkaManagementVersion,
   "com.lightbend.akka.discovery" %% "akka-discovery-kubernetes-api" % AkkaManagementVersion,
   "com.typesafe.akka" %% "akka-discovery" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-coordination" % AkkaVersion,
   // Common dependencies for logging and testing
   "com.typesafe.akka" %% "akka-slf4j" % AkkaVersion,
   "ch.qos.logback" % "logback-classic" % "1.2.3",
   "org.scalatest" %% "scalatest" % "3.1.2" % Test,
   // 2. Using gRPC and/or protobuf
   "com.typesafe.akka" %% "akka-http2-support" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-pki" % AkkaVersion,
   // 3. Using Akka Persistence
   "com.typesafe.akka" %% "akka-persistence-typed" % AkkaVersion,
   "com.typesafe.akka" %% "akka-serialization-jackson" % AkkaVersion,


ignasi35:
> Would be good with someones eye on it before I port it to the maven/java sample and apply to all steps of the guide.

looking ...

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
```diff
@@ -42,25 +42,37 @@ ThisBuild / dynverSeparator := "-"
...
   "com.lightbend.akka.discovery" %% "akka-discovery-kubernetes-api" % AkkaManagementVersion,
   "com.typesafe.akka" %% "akka-discovery" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-coordination" % AkkaVersion,
   // Common dependencies for logging and testing
   "com.typesafe.akka" %% "akka-slf4j" % AkkaVersion,
   "ch.qos.logback" % "logback-classic" % "1.2.3",
   "org.scalatest" %% "scalatest" % "3.1.2" % Test,
   // 2. Using gRPC and/or protobuf
   "com.typesafe.akka" %% "akka-http2-support" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-pki" % AkkaVersion,
```
ignasi35, line 75: This one looks weird. Why is `akka-pki` required? Hmmm, probably because of the mutuals TLS support on akka-cluster? We should move it to the top section (`// 1. Basic dependencies for a clustered app`).

ignasi35 (approved):
It's good to make the list of dependencies as explicit as possible but this is close to back-firing: "Hey, look! To build even a simple Akka Microservice you need all this huuuuuuge list of things!".

patriknw:
> before I port it to the maven/java sample

I don't think that will be needed since we use bom there now.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
johanandren: Yeap, transitive via `akka-remote` -> `akka-pki`. I'll move it up.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
```diff
@@ -42,25 +42,37 @@ ThisBuild / dynverSeparator := "-"
...
+  "com.typesafe.akka" %% "akka-actor" % AkkaVersion,
   "com.typesafe.akka" %% "akka-stream" % AkkaVersion,
   "com.typesafe.akka" %% "akka-cluster-typed" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-remote" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-distributed-data" % AkkaVersion,
   "com.typesafe.akka" %% "akka-cluster-sharding-typed" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster-sharding" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-cluster-tools" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-protobuf-v3_2.13" % AkkaVersion,
```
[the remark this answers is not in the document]
patriknw: I'm not sure I like this and if it is really needed. I can understand that we have to specify a few more leafs. Typical problem is akka-persistence-query because it only comes via the plugin or projections and not via any other core Akka dependency.

Theoretically it should work with only:
* akka-cluster-sharding-typed
* akka-persistence-query

(but I can have wrong understanding of how the sbt resolution works)

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
```diff
@@ -42,25 +42,37 @@ ThisBuild / dynverSeparator := "-"
...
   "com.typesafe.akka" %% "akka-actor-testkit-typed" % AkkaVersion % Test,
   "com.typesafe.akka" %% "akka-stream-testkit" % AkkaVersion % Test,
   // Akka Management powers Health Checks and Akka Cluster Bootstrapping
   "com.lightbend.akka.management" %% "akka-management" % AkkaManagementVersion,
   "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-http-core" % AkkaHttpVersion,
```
patriknw, line 61: this might be needed because akka-http is not using transitive dependencies at all

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
```diff
@@ -42,25 +42,37 @@ ThisBuild / dynverSeparator := "-"
...
   "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
+  "com.typesafe.akka" %% "akka-parsing" % AkkaHttpVersion,
   "com.lightbend.akka.management" %% "akka-management-cluster-http" % AkkaManagementVersion,
   "com.lightbend.akka.management" %% "akka-management-cluster-bootstrap" % AkkaManagementVersion,
   "com.lightbend.akka.discovery" %% "akka-discovery-kubernetes-api" % AkkaManagementVersion,
   "com.typesafe.akka" %% "akka-discovery" % AkkaVersion,
+  "com.typesafe.akka" %% "akka-coordination" % AkkaVersion,
```
patriknw, line 68: this shouldn't be needed since it's via akka-cluster

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
patriknw: it's an ordinary dependency from akka-remote so shouldn't be needed explicitly

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
johanandren: It is needed because akka-management modules will potentially pull in a different version of it (transitive from akka-http competing with akka-management)

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
johanandren: I think there's two aspects:

A. is the module _currently_ only pulled in as transitive to an Akka module
B. will the module _never_ be pulled in from some "third party" module (including not so third party such as `akka-management` or `akka-projections`. 

I bothered about neither here and thought better noisy/safe than sorry and listed _all_ Akka and Akka HTTP modules.

In the customer question where this popped up it was about Akka gRPC pulling in a different version of akka-http-core which at least I'd expect nobody to depend on directly, so I think it is a little bit dangerous to be "smart" about it with having a neat dependency list as the only benefit?

Do you think it is worth it?

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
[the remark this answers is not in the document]
johanandren: If we only list `akka-cluster-sharding-typed` that transitively pulls in `akka-cluster-sharding` which is also pulled in transitively by `akka-management-cluster-http` meaning that the highest two of those will "win" regardless of what the `AkkaVersion` is in this file. While if we pin them down, it should always work since we promise binary compatibility.

It's not a huge problem, but it is a problem we already saw someone have and that can be avoided by a little bit extra verbosity here.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
patriknw: For akka-http I think it's needed because it's not using transitive dependencies. Everything is provided.
I'd rather fix this for the problem that we have seen, which is Akka HTTP rather than adding all dependencies explicitly.

johanandren: There is a transitive depency on `akka-http-core` from `akka-http`, but in the concrete case Akka gRPC depended on a newer version of Akka HTTP so its transitive dependency won and they got a mixed set of Akka HTTP module versions.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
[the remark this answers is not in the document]
patriknw: ah, now I get it. akka-management-cluster-http would pull in a later version than the AkkaVersion in the build

patriknw (commented):
At first I didn't understand the reason why this was needed. This comment was enlightening https://github.com/akka/akka-platform-guide/pull/563#discussion_r573011419

octonato (commented):
Wouldn't that all be solved by running that plugin that checks that all Akka versions are aligned and if not fail the build? 

If we take that example with `akka-management-cluster-http`...

If it pulls an older version of `akka-cluster-sharding` that will be overwritten by `akka-cluster-sharding-typed` and we will stay in the `AkkaVersion` defined in this file. All fine.

If it pulls in a new version, the plugin will fail the build because we will have different versions of Akka. That also means we have a new version of Akka and that we probably want to update it.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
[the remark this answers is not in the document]
patriknw: However, thinking some more, it's actually dangerous to pin down the dependencies like this.
It would downgrade the Akka dependency to a lower version than what the 3rd library was built with. That can cause runtime errors, since we only provide backwards compatibility and not forward compatibility (we allow adding new API in patch versions).

For example. if akka-management-cluster-http was built with Akka 2.6.12 and the AkkaVersion here was pinned down to akka-cluster-sharding 2.6.10 it would use 2.6.10 even though akka-management-cluster-http requires 2.6.12.

Better to let the runtime version check detect that situation and the user must update the AkkaVersion to latest.

This is one of the reasons why we are conservative and not updating to the latest Akka version if not needed in libraries such as akka-management.

johanandren:
@octonato It wouldn't be solved by "that plugin" but it would make it noticeable earlier, currently, at least for core Akka modules it will fail any test because of the mismatch-check. This PR would avoid the problem altogether instead which I think is nicer, but at the cost of more verbose dependency listing.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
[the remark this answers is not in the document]
johanandren: Note though that since pinning down transitives is just making them explicit, that problem scenario with another dependency depending on newer which is not satisfied can also exists with the normal explicit dependencies.

Whatever we do, including nothing, has boths pros and cons. I personally do not feel any of them is great, but perhaps avoiding the `UnsatisfiedLinkError`-at-some-point-in-time and getting the fail-fast-module mismatch in _more_ cases is the best we can do.

So we leave the deps as they are?

octonato:
@johanandren, I thought we could configure the plugin to fail at compile time. 

In any case, I'm not sure it's nicer. We are increasing the noise in the build only to avoid that we get a build failure when versions get misaligned. 

I don't think users will be much impacted by this. They run the sample while going through the tutorial and they will get one that is already curated. Most usages will stop here.

If they start to change the dependencies and "break" it, they will run into the runtime plugin check anyway. So, their final experience will be the plugin errors.

johanandren:
@octonato we _already_ saw a customer getting into troubles and got stuck because of a module misalign caused by this, that's the original motivation for doing it

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-scala/build.sbt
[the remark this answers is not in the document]
patriknw: My point is that the pinning kind of disables the version mismatch check.

If the real problem you saw was because of library pulling in later akka or akka-http version than given in the app build I think we should leave it as it was. Especially if that problem was detected by the version check.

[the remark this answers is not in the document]
johanandren: Ok, let's close this PR then.

Akka HTTP has a check for supported Akka version, and some form of Akka HTTP version check (although that is not set up the same as Akka, so not sure it notices module mismatch https://github.com/akka/akka-http/blob/master/project/VersionGenerator.scala#L15 )

johanandren:
Closing after discussion with Patrik in https://github.com/akka/akka-platform-guide/pull/563#discussion_r572930399

Issue #577, opened 2021-02-15
IllegalStateException: Pool shutdown unexpectedly


patriknw:
This is harmless since it's only in the shutdown of tests but would be nice if we could avoid the noise

```
2021-02-15T09:48:54.9452923Z [ERROR] [02/15/2021 09:48:54.927] [IntegrationTest-akka.actor.default-dispatcher-3] [akka://IntegrationTest@127.0.0.1:38697/system/pool-master] connection pool for Pool(shared->http://127.0.0.1:40223) has shut down unexpectedly
2021-02-15T09:48:54.9454886Z java.lang.IllegalStateException: Pool shutdown unexpectedly
2021-02-15T09:48:54.9456282Z 	at akka.http.impl.engine.client.PoolInterface$Logic.postStop(PoolInterface.scala:206)
2021-02-15T09:48:54.9458171Z 	at akka.stream.impl.fusing.GraphInterpreter.finalizeStage(GraphInterpreter.scala:599)
2021-02-15T09:48:54.9460099Z 	at akka.stream.impl.fusing.GraphInterpreter.finish(GraphInterpreter.scala:324)
2021-02-15T09:48:54.9462114Z 	at akka.stream.impl.fusing.GraphInterpreterShell.tryAbort(ActorGraphInterpreter.scala:664)
2021-02-15T09:48:54.9464236Z 	at akka.stream.impl.fusing.ActorGraphInterpreter.$anonfun$postStop$1(ActorGraphInterpreter.scala:806)
2021-02-15T09:48:54.9466288Z 	at akka.stream.impl.fusing.ActorGraphInterpreter.$anonfun$postStop$1$adapted(ActorGraphInterpreter.scala:806)
2021-02-15T09:48:54.9467831Z 	at scala.collection.immutable.Set$Set1.foreach(Set.scala:168)
2021-02-15T09:48:54.9469464Z 	at akka.stream.impl.fusing.ActorGraphInterpreter.postStop(ActorGraphInterpreter.scala:806)
2021-02-15T09:48:54.9471310Z 	at akka.actor.Actor.aroundPostStop(Actor.scala:556)
2021-02-15T09:48:54.9472449Z 	at akka.actor.Actor.aroundPostStop$(Actor.scala:556)
2021-02-15T09:48:54.9474429Z 	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundPostStop(ActorGraphInterpreter.scala:691)
[...]

PR #575, opened 2021-02-15
remove guardian Behavior in Main

patriknw:
* Create the ActorSystem without a custom guardian Behavior.
* Initialize other things "from the outside"

(I will update Scala and all copies if we agree that this is the way to go)

PR #581, opened 2021-02-15
Register a connection pool shutdown task

ignasi35:
Not sure I'm on the right track. I'll copy/paste in other versions when approved.

References #577

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/Main.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/Main.java
@@ -33,13 +33,11 @@ public class Main {
 
     ShoppingCart.init(system);
 
-    ApplicationContext springContext =
-        SpringIntegration.applicationContext(system.settings().config());
+    ApplicationContext springContext = SpringIntegration.applicationContext(system);
+    JpaTransactionManager transactionManager = springContext.getBean(JpaTransactionManager.class);
 
     ItemPopularityRepository itemPopularityRepository =
         springContext.getBean(ItemPopularityRepository.class);
-
-    JpaTransactionManager transactionManager = springContext.getBean(JpaTransactionManager.class);
 
     ItemPopularityProjection.init(system, transactionManager, itemPopularityRepository);
 
--- a/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
@@ -1,8 +1,5 @@
 package shopping.cart.repository;
 
-import akka.Done;
-import akka.actor.CoordinatedShutdown;
-import akka.actor.typed.ActorSystem;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigValue;
 import com.zaxxer.hikari.HikariDataSource;
@@ -10,7 +7,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import javax.sql.DataSource;
 import org.springframework.context.annotation.Bean;
--- a/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
@@ -1,5 +1,8 @@
 package shopping.cart.repository;
 
+import akka.Done;
+import akka.actor.CoordinatedShutdown;
+import akka.actor.typed.ActorSystem;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigValue;
 import com.zaxxer.hikari.HikariDataSource;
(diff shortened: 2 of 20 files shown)


docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
```diff
@@ -57,6 +63,14 @@ public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
     // load additional properties from config jdbc-connection-settings.additional-properties
     factory.setJpaProperties(additionalProperties());
 
+    CoordinatedShutdown.get(system).addTask(
+            CoordinatedShutdown.PhaseActorSystemTerminate(),
```
patriknw, line 67: might be better to use PhaseBeforeActorSystemTerminate
documented as:
"Phase for custom application tasks that are to be run after cluster shutdown and before ActorSystem termination."

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
```diff
@@ -57,6 +63,14 @@ public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
...
+            "shutdown-spring",
+            () -> {
+              factory.destroy();
```
patriknw, line 70: There is one risk of using CoordinatedShutdown here. Spring resources would be destroyed before ordinary actors are stopped so the actors can't use them at the end.

In the example it's not a problem because entities and projections are stopped before this phase.

Might anyway be better to destroy after the ActorSystem. By the way, perhaps we should add another phase after system termination to CoordinatedShutdown so that it can be used in the same way as other phases instead of ActorSystem.whenTerminated.

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
```diff
@@ -57,6 +63,14 @@ public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
     // load additional properties from config jdbc-connection-settings.additional-properties
     factory.setJpaProperties(additionalProperties());
 
+    CoordinatedShutdown.get(system).addTask(
```
patriknw, line 66: Could we close the Spring ApplicationContext (from Main) instead of this specific bean?

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
ignasi35: My intention was to only shutdown the JpsTransaction Manager (I used a confusing task name) but you raise a good point. The spring context is created in parallel to the actor system not _by_ the actor system.

I just realised there may be a JVM Shutdown hook for spring out of the box that we would have to disable (would have to look into it).

I think we need to `    system.getWhenTerminated().thenAccept(done -> factory.destroy());`

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
```diff
@@ -57,6 +63,14 @@ public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
...
+            CoordinatedShutdown.PhaseActorSystemTerminate(),
+            "shutdown-spring",
+            () -> {
+              factory.destroy();
+              return CompletableFuture.completedFuture(Done.getInstance());
+            });
```
ignasi35, line 72:
```suggestion
    system.getWhenTerminated().thenAccept(done -> factory.destroy());
```

ignasi35: Actually, a wider shutdown invocation over the whole spring context would be better:

```
// In Main.java
    system.getWhenTerminated().thenAccept(done -> ((AnnotationConfigApplicationContext)springContext).close());
```

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/repository/SpringConfig.java
ignasi35: just saw this 🤦🏼‍♂️

docs-source/docs/modules/microservices-tutorial/examples/05-shopping-cart-service-java/src/main/java/shopping/cart/Main.java
```diff
@@ -55,6 +63,9 @@ public Main(ActorContext<Void> context) {
     PublishEventsProjection.init(system, transactionManager);
     // end::PublishEventsProjection[]
 
+
+    system.getWhenTerminated().thenAccept(done -> ((AnnotationConfigApplicationContext)springContext).close());
```
patriknw, line 67: can be moved up, where the ApplicationContext is created, or even inside `SpringIntegration.applicationContext`

patriknw (commented):
I like that better

ignasi35:
> I like that better

Yeah, it looks cleaner.

ignasi35:
If I get an approval I'll replicate the changes in all services and versions of the service.

patriknw (commented):
looking good, but before you change everywhere we must have #575 merged, and I don't know what's going on with the tests

patriknw:
@ignasi35 the other PR has been merged now

patriknw (approved):
LGTM

PR #582, opened 2021-02-16
[Proposal] Keep external dependencies on the edge

ignasi35:
Built on top of #575

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
@@ -1,16 +1,13 @@
 package shopping.cart;
 
 import akka.actor.typed.ActorSystem;
-import akka.actor.typed.Behavior;
-import akka.actor.typed.javadsl.AbstractBehavior;
-import akka.actor.typed.javadsl.ActorContext;
 import akka.actor.typed.javadsl.Behaviors;
-import akka.actor.typed.javadsl.Receive;
 // tag::SendOrderProjection[]
 import akka.grpc.GrpcClientSettings;
 // end::SendOrderProjection[]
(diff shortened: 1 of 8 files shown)


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/main/java/shopping/cart/Main.java
```diff
@@ -35,15 +38,13 @@ public static void init(ActorSystem<Void> system, ShoppingOrderService orderServ
     ApplicationContext springContext =
             SpringIntegration.applicationContext(system.settings().config());
     JpaTransactionManager transactionManager = springContext.getBean(JpaTransactionManager.class);
-
     ItemPopularityRepository itemPopularityRepository =
             springContext.getBean(ItemPopularityRepository.class);
+    Supplier<HibernateJdbcSession> sessionSupplier = () -> new HibernateJdbcSession(transactionManager);
```
patriknw, line 43: I'm not convinced (yet) that this indirection makes it easier to understand?

ignasi35: I think it's a matter of wording in the accopanying docs. An Akka Projection needs a `Supplier<JdbcSession>` (even if that's not how it's presented in the docs https://doc.akka.io/docs/akka-projection/current/jdbc.html).

In the Akka Projection docs we just add an argument `() -> sessionProvider.newInstance()` but that's a `Supplier<JdbcSession>`. 

Alternatively, I would copy the `HibernateSessionFactory` from the Akka projections docs into this code base.

-----

My initial intention was to push spring dependencies to the edge, but I think I have an extra motivation now which is bring this code closer to the samples in the Akka Projection docs.

ignasi35:
I've had this on hold for a while. Closing. We can revisit it again in the future.

PR #583, opened 2021-02-16
Docs for Kubernetes Lease for SBR

ignasi35:


docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,57 @@
...
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  image: <image>
+  splitBrainResolver:
+    lease:
+      provider: "kubernetes" # <1>
+        actorSystemName: "ShoppingCartService" # <2>
```
ignasi35, line 53: The gist of it is that the CRD would include the following:

```
apiVersion: akka.lightbend.com/v1
kind: AkkaMicroservice
metadata:
  name: shopping-cart-service
spec:
  image: <image>
  splitBrainResolver: # 1
    lease: # 2
      provider: "kubernetes" 
        actorSystemName: "ShoppingCartService" # 3
```
1. a new section for all-things SBR
2. a subsection to tune a lease for SBR. At the moment we’d only support `"kubernetes"`but other leases could be included later. The operator may or may not take action depending on the provider.
3. parameters for the k8s lease provider (the minimal is the `ActorSystem` name)

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,57 @@
...
++
+.build.sbt:
+[source,scala,indent=0]
+----
+include::example$shopping-cart-service-scala/build.sbt[tag=akka-sbr-kubernetes-lease]
+----
+
+== Prepare your application
+
+When using a Kubernetes-backed Lease to improve the SBR decision process, you must make sure the name of your `ActorSystem` is the same for all the nodes of the Akka Cluster. At same time, you must make sure that different Akka Clusters deployed in the same Kubernetes cluster use different names. Each Akka Cluster must have a single Lease, and each Lease is identified by the name of the `ActorSystem`.
```
patriknw, line 33: same ActorSystem name doesn't have to be mentioned here, that is a requirement for Akka Cluster itself.

patriknw, line 33: Is it "Kubernetes cluster" or "Kubernetes namespace"? Is the CR global or scoped by namespace?

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
patriknw: Is there a config to define the lease name instead of using the ActorSystem name? Then we could use the Microservice name instead (`shopping-cart-service`) and inject that config via main.conf.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,57 @@
...
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  image: <image>
+  splitBrainResolver:
+    lease:
+      provider: "kubernetes" # <1>
```
patriknw, line 52: since we don't have any plans for other providers this seems unnecessary

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: I didn't find any setting to tune the name. Looking at the [docs](https://doc.akka.io/docs/akka-management/current/kubernetes-lease.html) and the [code](https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/sbr/SplitBrainResolverProvider.scala#L56) I thnk it's not possible at the moment.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: Yeah, that's the step I was least convinced too. Also, because this is a CR (that is a _kubernetes thingy_) even if we had other leases it makes sense that the CR configures the kubernetes' Lease. I'll simplify this step.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: With the section on setting up the CR, this section on tuning the Actor system name no longer makes sense.

ignasi35: > Is the CR global or scoped by namespace?

I think each `Lease` CR is namespaced.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
patriknw: then I think we should add that, we wanted a new release anyway

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
patriknw: good, then there can't be any name conflicts if we are using the AkkaMicroservice service name

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,51 @@
...
+metadata:
+  name: shopping-cart-service
+spec:
+  image: <image>
+  splitBrainResolver:
+    useKubernetesLease: true # <1>
+
+----
+
+<1> When using a kubernetes lease as a backend for the Split-Brain Resolver, the poerator will make sure the Lease exists and will name it `<metadata.name>-akka-sbr`. In this case, the lease will be `shopping-cart-service-akka-sbr`.
```
ignasi35, line 51: This naming strategy will require https://github.com/akka/akka/pull/30049

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,51 @@
...
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+spec:
+  image: <image>
+  splitBrainResolver:
+    useKubernetesLease: true # <1>
```
chbatey, line 47: Good to have it explicit, I don't think we should configure it automatically

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
```diff
@@ -0,0 +1,51 @@
...
++
+.build.sbt:
+[source,scala,indent=0]
+----
+include::example$shopping-cart-service-scala/build.sbt[tag=akka-sbr-kubernetes-lease]
+----
+
+== Setup you Kubernetes Cluster and AkkaMicroservice
+
+When xref:deployment:aws-install.adoc#_install_akka_operator[installing the Akka Operator] on your Kubernetes cluster using the Helm chart you will also install the Akka Kubernetes Lease CRD.
```
chbatey, line 33: Is the helm chart going to do it or do we need instructions or a pointer to the akka management docs for how to do it?

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
chbatey, line 51: typo: poerator

chbatey: Good to use metadata.name as it is already known to be a valid k8s name. Might get too long tho

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: My current implementation is that the Helm chart installs the Lease CRD and the operator stops running if the Lease CRD is removed. I'm treating the Lease CRD with the same importance than the AkkaMicroservices one.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: Ah, good catch. I'll add a check and a test.

When the lease name is too long the operator should fail.

Question is how to fix the issue. I see a few options: 

1. require a name override in `spec.splitBrainResolver.leaseName`, 
2. require a short name for the service in `metadata.shortName`, 
3. synthesize a new name abbreviating the service name, e.g. `shop-cart-svc-akka-sbr`

I'm leaning towards (2).

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
ignasi35: > configure it automatically

Could we? Users need to add the dependency.

Note I'm adding a few setting into `main.conf`, though. My idea is that _only_  adding the dependency on the build and doing the explicit enable on the CR it would work using the default settings.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
patriknw: we have similar problem for several other resources derived from the CR name, see https://github.com/lightbend/akka-platform-operator/issues/437

Let's discuss solution there.

docs-source/docs/modules/how-to/pages/sbr-kubernetes-lease.adoc
patriknw: > the operator stops running if the Lease CRD is removed

I can't see that in the implementation? I think that would be wrong thing to do. It's an optional feature and if someone would like to not use it and remove the Lease CRD that shouldn't impact the operator.

It's convenient that helm installs the Lease CRD.

docs-source/docs/modules/how-to/examples/shopping-cart-service-scala/build.sbt
```diff
@@ -84,6 +84,11 @@ libraryDependencies ++= Seq(
   "com.typesafe.akka" %% "akka-stream-kafka" % AlpakkaKafkaVersion,
   "com.lightbend.akka" %% "akka-projection-testkit" % AkkaProjectionVersion % Test,
 
+  // 5. Kubernetes Lease (used by SBR)
```
ignasi35, line 87: I added this only on the sample app used for snippets of the how-to, but I think it should be included in all code bases of the tutorial. Same way the tutorial code bases include `"akka-discovery-kubernetes-api"`.

RayRoestenburg (approved):

PR #584, opened 2021-02-18
Initial GKE ingress docs

chbatey:


docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
```diff
@@ -0,0 +1,75 @@
...
+.kubernetes/shopping-cart-service-cr.yml:
+[source,yaml]
+----
+apiVersion: akka.lightbend.com/v1
+kind: AkkaMicroservice
+metadata:
+  name: shopping-cart-service
+  namespace: "shopping"
+  annotations:
+    akka.lightbend.com/ingress-class: gce
```
patriknw, line 50: Do I understand it correctly that enabling `grpcIngress` without also specifying this annotation (gce or alb) will not work? When we talked about annotations I thought it would be used for custom options, but we could still support the default without them?

Could we add the ingress-class as a property fo the `grpcIngress` section instead so that the default ingress can be enabled at one place?

patriknw: what happens if both grpcIngress and httpIngress are enabled and they need different annotations? I thought the support for annotations would be in grpcIngress/httpIngress sections rather than at the top level?

chbatey: > what happens if both grpcIngress and httpIngress are enabled and they need different annotations? I thought the support for annotations would be in grpcIngress/httpIngress sections rather than at the top level?

We could have separate annotations ones for http and grpc or put it as a property as you say. Annotations only work in the metadata section. 

If you don't specify any annotations you get a normal Ingress/Service that would work with any Ingress Controller that doesn't require any annotations on the Ingress resource e.g. Minikube Ingress would work. 

Given that you have to install the AWS ingress controller manually that made me think that we shouldn't just default based on the type of cluster we are in.

patriknw: I think we should make the choice (ingress-class) more explicit by making it an explicit property in the grpcIngress/httpIngress sections. We can still have it optional as you say.

I think I misunderstood what we would use annotations for. I thought that was for custom options (AWS LoadBalancer Controller has [massive amount of annotations](https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/) for various things). The operator would just transparently propagate them from the CR to the created Service or Ingress. That is a different story.

chbatey: > I think we should make the choice (ingress-class) more explicit by making it an explicit property in the grpcIngress/httpIngress sections. We can still have it optional as you say.

Agreed.

> I think I misunderstood what we would use annotations for. I thought that was for custom options (AWS LoadBalancer Controller has massive amount of annotations for various things).

I do use one like that, the gke.akka.lightend.io/backend-config but it is for our behavior, not just forwaded on.

> I thought that was for custom options (AWS LoadBalancer Controller has massive amount of annotations for various things). The operator would just transparently propagate them from the CR to the created Service or Ingress. That is a different story.

I didn't get to that but do agree it will be useful and we need to allow users to specify which annotations go where some how

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
```diff
@@ -0,0 +1,75 @@
+= AWS LoadBalancer Controller
```
debasishg, line 1: shouldn't this be GKE LoadBalancer Controller ?

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
patriknw: I created a new issue for annotations https://github.com/lightbend/akka-platform-operator/issues/477

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
chbatey: Yep, or just "GKE Ingress" as you aren't aware what the controller is called

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
```diff
@@ -0,0 +1,79 @@
+= GKE Ingress
+:page-toclevels: 3
+
+include::partial$include.adoc[]
+
+gRPC and HTTP ingress with the https://cloud.google.com/kubernetes-engine/docs/concepts/ingress[GKE Ingress Controller {tab-icon}, window="tab"] is supported
+ion documentation {tab-icon}, window="tab"]
```
ignasi35, line 7:
```suggestion
gRPC and HTTP ingress with the https://cloud.google.com/kubernetes-engine/docs/concepts/ingress[GKE Ingress Controller {tab-icon}, window="tab"] is supported  <something seems to be missing here>
<and maybe here too> ion documentation {tab-icon}, window="tab"]
```

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
```diff
@@ -0,0 +1,79 @@
...
+  name: shopping-cart-service
+  namespace: "shopping"
+  annotations:
+spec:
+  image: <image>
+  grpcIngress:
+    enabled: true
+    certificate: "<secret-name>"
+    class: "gce"
+----
```
ignasi35, line 56:
```suggestion
----
kubectl create secret tls my-ingress-secret --cert dummy-cert.pem  --key dummy-key.pem
----

This is referenced from the Akka Microservice.

== Enable grpcIngress

Add the `grpcIngres` section to the deployment descriptor with class set to `gce`:

.kubernetes/shopping-cart-service-cr.yml:
[source,yaml]
----
apiVersion: akka.lightbend.com/v1
kind: AkkaMicroservice
metadata:
  name: shopping-cart-service
  namespace: "shopping"
  annotations:
spec:
  image: <image>
  grpcIngress:
    enabled: true
    certificate: "my-ingress-secret"
    class: "gce"
----
```

docs-source/docs/modules/deployment/pages/gcp-ingress.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+
+When the deployment descriptor has been applied the Akka Operator will create:
+
+* An Ingress with the appropriate GKE annotations
+* A `ClusterIP` service.
+* A `BackendConfig' to configure a health check to match the readiness check
```
ignasi35, line 63:
```suggestion
* an `Ingress` with the appropriate GKE annotations
* a `Service` of type `ClusterIP`
* a https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#associating_backendconfig_with_your_ingress[`BackendConfig` {tab-icon}, window="tab"] to configure a health check to match the readiness check
```

ignasi35 (changes requested):
Left a few comments.

ignasi35:
Do we require a [minimum GKE version](https://github.com/lightbend/akka-platform-operator/pull/476#discussion_r580629397)?

chbatey:
TODO: add a note about min GKE version

ignasi35:
> TODO: add a note about min GKE version

I think this is no longer necessary. The Stable release channel is already on 1.17 (https://cloud.google.com/kubernetes-engine/docs/release-notes)

ignasi35:
@debasishg I've pushed a couple of commits and I think this is ready for a fainl review.

Note that I rewrote the `TLS certificate` section as if the `httpIngress` section already supported TLS (ref https://github.com/lightbend/akka-platform-operator/issues/477).

ignasi35 (approved):

ignasi35:
As discussed on the team call this is ready to merge.

Issue #578, opened 2021-02-15
failed: ItemPopularityIntegrationTest: expected:<9> but was:<5>


patriknw:
I thought we had fixed this in https://github.com/akka/akka-platform-guide/issues/534 but it's still failing

```
21-02-15T09:51:05.8458079Z Failed tests:   consumeCartEventsAndUpdatePopularityCount(shopping.cart.ItemPopularityIntegrationTest): expected:<9> but was:<5>
```

ignasi35:
I suspect we are running tests in parallel and the actor system from `ItemPopularityIntegrationTest` is joining the cluster from `IntegrationTest` or is shutdown when that cluster shutsdown.

see 

```
2021-02-15T09:50:47.7787540Z [2021-02-15 09:50:47,777] [INFO] [akka.management.scaladsl.AkkaManagement] [akka://IntegrationTest@127.0.0.1:38819] [] [IntegrationTest-akka.actor.default-dispatcher-6] - Binding Akka Management (HTTP) endpoint to: 127.0.0.1:34265
2021-02-15T09:50:47.7798860Z [2021-02-15 09:50:47,779] [INFO] [akka.management.scaladsl.AkkaManagement] [akka://IntegrationTest@127.0.0.1:36589] [] [IntegrationTest-akka.actor.default-dispatcher-6] - Binding Akka Management (HTTP) endpoint to: 127.0.0.1:41427
2021-02-15T09:50:47.7858680Z [2021-02-15 09:50:47,785] [INFO] [akka.management.scaladsl.AkkaManagement] [akka://IntegrationTest@127.0.0.1:41181] [] [IntegrationTest-akka.actor.default-dispatcher-3] - Binding Akka Management (HTTP) endpoint to: 127.0.0.1:34395
```

patriknw:
I messed up and took the output from wrong test, and now I can't find the failed build anymore. We'll have to wait until it happens again.

ignasi35:
Seen it again:

```
...
[2021-02-25 11:56:54,313] [INFO] [com.zaxxer.hikari.HikariDataSource] [] [] [ItemPopularityIntegrationTest-akka.actor.default-dispatcher-20] - db - Shutdown initiated...
[2021-02-25 11:56:54,321] [INFO] [com.zaxxer.hikari.HikariDataSource] [] [] [ItemPopularityIntegrationTest-akka.actor.default-dispatcher-20] - db - Shutdown completed.
[...]

PR #589, opened 2021-02-23
Close gRPC client in tests

ignasi35:
References #577

Pending: 

- [x] replicate to scala code base, 
- [x] replicate to other tests in the same service and other services in the microservices tutorial and other modules (e.g. `how-to/`), 
- [x] copy/paste code to all versions of the sample code.


ignasi35:
This continually fails on #578 and #591

patriknw (commented):
good

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/Main.java
```diff
@@ -65,6 +66,14 @@ static ShoppingOrderService orderServiceClient(ActorSystem<?> system) {
...
             .withTls(false);
 
-    return ShoppingOrderServiceClient.create(orderServiceClientSettings, system);
+    ShoppingOrderServiceClient client =
+        ShoppingOrderServiceClient.create(orderServiceClientSettings, system);
+    CoordinatedShutdown.get(system)
+        .addTask(
+            CoordinatedShutdown.PhaseBeforeServiceUnbind(),
+            "close-test-client-for-grpc",
+            () -> client.close());
```
patriknw, line 75: Is this needed? Does it solve anything over letting it be shutdown when the ActorSystem terminates? Is it "Pool shutdown unexpectedly" here also, not only in tests?

The reason I'm asking is that I don't want us to more infra structure code than necessary?

Could it be a possible improvement in Akka HTTP or gRCP to shutdown more clean on ordinary ActorSystem termination?

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/Main.java
```diff
@@ -65,6 +66,14 @@ static ShoppingOrderService orderServiceClient(ActorSystem<?> system) {
                 system)
             .withTls(false);
 
-    return ShoppingOrderServiceClient.create(orderServiceClientSettings, system);
+    ShoppingOrderServiceClient client =
+        ShoppingOrderServiceClient.create(orderServiceClientSettings, system);
+    CoordinatedShutdown.get(system)
+        .addTask(
+            CoordinatedShutdown.PhaseBeforeServiceUnbind(),
```
patriknw, line 73: This phase might be too early. Client might sill be in used, projections haven't been stopped yet.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/test/java/shopping/cart/IntegrationTest.java
```diff
@@ -107,6 +108,11 @@ public TestNodeFixture(int grcpPort, List<Integer> managementPorts, int manageme
     public shopping.cart.proto.ShoppingCartService getClient() {
       if (client == null) {
         client = shopping.cart.proto.ShoppingCartServiceClient.create(clientSettings, system);
+        CoordinatedShutdown.get(system)
+            .addTask(
+                CoordinatedShutdown.PhaseBeforeServiceUnbind(),
```
patriknw, line 113: isn't it better to close it as late as possible?

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/Main.java
ignasi35: I was on cruise control but I think this not right. This gRPC client is part of a projection that will be stopped when leaving the cluster. I'll reset this change.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/Main.java
ignasi35: Yeah. I've stepped back. I think that as long as this client is only used in the projection it's OK if we don't register a task to shutdown the client.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/test/java/shopping/cart/IntegrationTest.java
ignasi35: I'm not sure. My thinking was:

 - if the server unbinds before the client has shut down, any open connection will be dropped

patriknw: ok, this is the client that is used by test itself. good to close that as early as possible.

patriknw (approved):
LGTM

PR #594, opened 2021-02-25
Use non-primitive type for the Version field

ignasi35:
References #578 

From https://docs.spring.io/spring-data/jpa/docs/2.4.5/reference/html/#jpa.entity-persistence.saving-entites.strategies

> Version-Property and Id-Property inspection (default): By default Spring Data JPA inspects
>  first if there is a Version-property of non-primitive type. If there is, the entity is
>  considered new if the value of that property is null. Without such a Version-property Spring
>  Data JPA inspects the identifier property of the given entity. If the identifier property is
>  null, then the entity is assumed to be new. Otherwise, it is assumed to be not new.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
@@ -11,12 +11,13 @@ public class ItemPopularity {
   @Id private final String itemId;
 
   // optimistic locking
-  @Version private final long version;
+  @Version private final Long version;
 
   private final long count;
 
   public ItemPopularity() {
-    this.version = 0;
+    // null version means the entity is not on the DB
+    this.version = null;
     this.itemId = "";
     this.count = 0;
   }
--- a/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
@@ -44,9 +44,4 @@ public class ItemPopularity {
     return new ItemPopularity(itemId, version, count + delta);
   }
 
-  @Transient
-  @JsonIgnore
-  boolean isNew() {
-    return itemId.equals("") || version <= 0;
-  }
 }
--- a/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
+++ b/docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
@@ -1,6 +1,5 @@
 package shopping.cart;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import javax.persistence.*;
 
 @Entity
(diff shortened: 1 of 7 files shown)


docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/test/java/shopping/cart/ItemPopularityIntegrationTest.java
```diff
@@ -111,4 +111,50 @@ public void consumeCartEventsAndUpdatePopularityCount() throws Exception {
...
+
+    // ... when 99 concurrent carts add `item1`...
+    for (int i = 1; i < cartCount; i++) {
+      System.out.print(".");
+      CompletionStage<ShoppingCart.Summary> rep =
+          sharding
+              .entityRefFor(ShoppingCart.ENTITY_KEY, "concurrent-cart" + i)
+              .askWithStatus(
+                  replyTo -> new ShoppingCart.AddItem(item, itemCount, replyTo), timeout);
+      rep.toCompletableFuture().get(3, SECONDS);
```
patriknw, line 147: Would be more concurrent if not waiting here, but collecting the result and wait on them afterwards. No need to wait at all since the awaitAssert below should cover that.

docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
```diff
@@ -11,12 +11,13 @@
   @Id private final String itemId;
 
   // optimistic locking
-  @Version private final long version;
+  @Version private final Long version;
```
patriknw, line 14: Strange that it wouldn't understand long 0 as the same thing since first version is 1 (I guess).

patriknw (commented):
looking good

docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java/src/main/java/shopping/cart/ItemPopularity.java
patriknw: note that we already have a `boolean isNew` below. Can that be removed now?

ignasi35: I think the problem, then, is that `isNew` is implemented but the class doesn't implement `Persistable<>`:

> Implementing Persistable: If an entity implements Persistable, Spring Data JPA delegates 
> the new detection to the isNew(…) method of the entity. See the JavaDoc for details.

I'll delete `isNew`.

ignasi35: > I'll delete isNew.

Or was that a purposeful choice? I mean, prefer using `isNew` with a primitive type for `@Version` (fewer autoboxing and allocations).

ignasi35: I've removed `isNew`

patriknw: no purpose, good with `Long` if that is what JPA is using by default

ignasi35:
"'Cause if you liked it, then you should have put a ✅ on it
If you liked it, then you shoulda put a ✅ on it
Don't be mad once you see that he want it
'Cause if you liked it, then you shoulda put a ✅ on it
Oh, oh, oh, oh, oh, oh."

patriknw (approved):
LGTM

Issue #591, opened 2021-02-24
Failed: shopping.cart.IntegrationTest - expected:<42> but was:<84>


johanandren:
https://github.com/akka/akka-platform-guide/pull/590/checks?check_run_id=1968707991

```
updateAndProjectFromDifferentNodesViaGrpc(shopping.cart.IntegrationTest)  Time elapsed: 9.189 sec  <<< FAILURE!
788
java.lang.AssertionError: expected:<42> but was:<84>
789
	at org.junit.Assert.fail(Assert.java:89)
790
	at org.junit.Assert.failNotEquals(Assert.java:835)
791
	at org.junit.Assert.assertEquals(Assert.java:647)
792
	at org.junit.Assert.assertEquals(Assert.java:633)
793
	at shopping.cart.IntegrationTest.lambda$updateAndProjectFromDifferentNodesViaGrpc$e3b33215$1(IntegrationTest.java:289)
```

johanandren:
Same on the Scala side:

```
[info] - should update and project from different nodes via gRPC *** FAILED *** (13 seconds, 402 milliseconds)
670
[info]   The code passed to eventually never returned normally. Attempted 50 times over 10.056724471 seconds. Last failure message: 84 did not equal 42. (IntegrationSpec.scala:286)
671
[info]   org.scalatest.exceptions.TestFailedDueToTimeoutException:
672
[info]   at org.scalatest.enablers.Retrying$$anon$4.tryTryAgain$2(Retrying.scala:185)
673
[info]   at org.scalatest.enablers.Retrying$$anon$4.retry(Retrying.scala:192)
674
[info]   at org.scalatest.concurrent.Eventually.eventually(Eventually.scala:402)
675
[info]   at org.scalatest.concurrent.Eventually.eventually$(Eventually.scala:401)
676
[info]   at shopping.cart.IntegrationSpec.eventually(IntegrationSpec.scala:135)
677
[info]   at shopping.cart.IntegrationSpec.$anonfun$new$6(IntegrationSpec.scala:286)
```

patriknw:
Probably still something with bootstrap performing double self join. See https://github.com/akka/akka-platform-guide/issues/586

ignasi35:
This hasn't failed in a while...

PR #595, opened 2021-02-25
Prevent multiple self-join in cassandra sample apps

ignasi35:
References #591 

In #586 we forgot about the example apps in `modules/how-to`.


ignasi35:
Failure is #594

patriknw (approved):
LGTM, good catch

Issue #593, opened 2021-02-25
Reorganize the Cloud deployment section and introductions


ignasi35:
The currently deployed guide has a bit of a mix on the Cloud Deployment section:

<img width="313" alt="Screenshot 2021-02-25 at 12 10 31" src="https://user-images.githubusercontent.com/762126/109145285-97ab0500-7762-11eb-9fec-f04a39e55645.png">

A more convenient/usable distribution of the information could be:

- (some intro here?)
- AWS
	- EKS
	- RDS
	- Ingress thingies
	- TLS
	- another AWS-specific tech
	- ...
- GCP 
	- GCP k8s cluster
	- GCP Cloud SQL
	- Ingress thingies
	- TLS
	- another GCP-specific tech
	- ...
- Managed Services
	- Kafka - Confluent Cloud
	- Kafka - Strimzi Operator
	- Datastax Astra (?)
	- other managed services
- Akka Operator Reference
- Troubleshooting

ignasi35:
PS: I tried to follow the exact same list on the AWS and GCP sections:

- cluster
- SQL
- ingress
- TLS
- others...

ignasi35:
ping @thothmoses

thothmoses:
@ignasi35, @rstento I took a swipe at outlining this so we can begin rolling forward on it. This should happen in parallel with the creation of the additional GPC materials.
https://docs.google.com/document/d/1Jh9VMfrOKkaXgj4EP4NTeTG7KRPXc6Cu24rEz_vWLaQ/edit?usp=sharing

patriknw:
Keep me in the loop of this. I'm not sure I understand the suggestions in the document.

rstento:
@thothmoses, nice job! It took me a while to grok what you were suggesting and I've countered in a number of places. I do believe that re-organizing the whole guide at this point is just too much to bite off. I recommend instead focusing on what are the two most important pieces: Rewritting the first page and reorganizing the Deployment chapter in prep for including GCP content. @patriknw @ignasi35 @jtownley, would you agree? In fact, this issue is titled "Reorganize deployment section" and I think that can be started. Another issue should be created for reorganizing the guide and one for rewriting the Introduction (first page). These related issues can be organized under an epic.

rstento:
Also I wanted to note that as @thothmoses works through reorganizing these pages, it is very likely that he will see ways of fine-tuning, such as renaming pages and headings and shuffling them around.

patriknw:
[...]

PR #598, opened 2021-02-26
Proposal for deployment restructure

ignasi35:
References #593

the reasoning is that if I'm a GCP user, I don't want to see anything that's AWS specific.

The proposal looks like this:

Screenshot_2021-02-26_at_14_33_58

Note how the section on "Integrations with Databases and message brokers" includes _all_ DBs and brokers even if that means duplicating links to the `aws-rds.adoc` or `gcp-sql.adoc`.


ignasi35:
This task was picked up by @thothmoses so I will close this PR awaiting for his work on a new branch.

PR #600, opened 2021-02-26
GCP user instructions

ignasi35:
References #387

Drafted some initial steps. I've left a handful of TODOs and probably skipped alternatives, I'm not familiar enough with all the options. It'd be better if someone with in-depth knowledge on the installer can pick this PR from this point on.

### DIFF
--- a/docs-source/docs/modules/deployment/pages/gcp-install.adoc
+++ b/docs-source/docs/modules/deployment/pages/gcp-install.adoc
@@ -1,0 +1,117 @@
+= Installation in Google Cloud Platform
+:page-toclevels: 3
+
+include::partial$include.adoc[]
+
+To install Akka Cloud Platform on Google Kubernetes Engine (GKE), you must have a Google Cloud Platform account and a project.
+
+:sectnums:
+== Verify Prerequisites
+
+To install and use the Akka Cloud Platform, you must have the following tools installed. We recommend using the latest versions of these tools. If you do not have them yet or need to update, we provide links to installation instructions:
+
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/deployment/pages/gcp-install.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+. GKE cluster where you are installing the operator (or use the "Create a new cluster" is you still don't have one)
+. Namespace <-- TODO: explain what the recommended value and implications of using a dedicated namespace vs `default`
+. App instance name: select a unique name. We recommend using the default or `akka-platform-operator-1`.
+. akka-operator.serviceAccount.name <-- TODO: explain this
+. Reporting Service Account <-- TODO: explain this
+
+Finally, take a moment to review all the permissions the Akka Clout Platform operator will use in order to operate your applications.
+
+== TODO: install from helm?
```
andreaTP, line 108: To have metering and billing enabled we don't have this option (e.g. Marketplace needs to inject values that are not public and we can't share with the users)

andreaTP (changes requested):
I made a few suggestions! Thanks a lot @ignasi35 for taking over!

ignasi35:
@andreaTP as discussed offline. I agree that maybe we can simplify this and just make the GCP instructions over the bare minimum (using UI) to get up and running. Then, on future PR's, we can improve the docs to cover cases where the user prefers using CLI, or they have a pre-existing cluster,....

ignasi35:
LGTM

andreaTP (approved):

RayRoestenburg (approved):

PR #605, opened 2021-03-02
Update ShoppingCartService.proto

staniopolis:
We use "count" in sql queries

References #xxxx


andreaTP:
Thanks for your PR, unfortunately, I don't understand the motivation behind it.
Please feel free to re-open and give more context!

PR #611, opened 2021-03-04
GCP user instructions

ignasi35:
Revert #609 so we get the instructions back.


ignasi35:
As discussed on the team call, this is no longer blocked.

RayRoestenburg (approved):

Issue #52, opened 2020-08-20
how to: write an integration test


patriknw:
The tutorial will probably only cover some unit tests. Integration testing with real Cassandra and Kafka can be a separate "how to recipe"

PR #624, opened 2021-03-12
Refactor settings loading on tests

ignasi35:
While working on #597 I struggled with some integration tests and how they were loading settings. 

(edit: this PR differs from #597 in that this PR assumes the start of Bootstrap and Management from settings and then goes to suggest a refactor on how settings are organized in tests)

Integration tests _must_  use random ports for gRPC, artery and akka management so those settings must be dynamic. On top of that, an integration test must start a 3-node (or 3 actorSystem) cluster  while another needs only 1 node.

Related to the concerns about ports above, tests need some generic overrides (e.g. persistence) and then some per-test overrides. 

Therefore, I am introducing an idiom to produce the `Config` each actor system on a test may need:

```scala
 private val config: Config =
      DynamicTestConfig.endpointConfig(serviceName, grpcPort) 
        .withFallback(DynamicTestConfig.clusteringConfig(serviceName, managementPorts, managementPortIndex)) 
        .withFallback(ConfigFactory.parseString(integrationSpecOverrides)) 
        .withFallback(ConfigFactory.parseResources("persistence-test.conf")) 
        .withFallback(ConfigFactory.load("application.conf")) 
        .resolve()
```

----------------

Setting as DRAFT until #597 is merged and then this is agreed and replicated to all code bases in the repo


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/resources/persistence-test.conf
```diff
@@ -1,3 +1,5 @@
+include "persistence"
```
ignasi35, line 1: A test that needs persistence should only need to load `persistence-test.conf` and things would Just Work ™️

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
```diff
@@ -0,0 +1,106 @@
...
+                       managementPorts: Seq[Int],
+                       managementPortIndex: Int): Config =
+    ConfigFactory.parseString(forCluster(serviceName, managementPorts, managementPortIndex))
+
+
+  // This can be extended with TLS ports, HTTP, etc.. and any port
+  // binding open for blocking calls
+  def endpointConfig(serviceName: String,
+                     grpcPort: Int): Config =
+    ConfigFactory.parseString(forGrpc(serviceName, grpcPort))
```
ignasi35, line 104: Build dynamic configs for cluster formation, akka management, and HTTP (not shown) and gRPC endpoints.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
```diff
@@ -36,39 +36,40 @@ import shopping.order.proto.OrderResponse
 import shopping.order.proto.ShoppingOrderService
 
 object IntegrationSpec {
-  val sharedConfig: Config = ConfigFactory.load("integration-test.conf")
```
ignasi35, line 39: Don't use per test conf files. Instead, move all the Config-creation logic to the test codebase.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
```diff
@@ -36,39 +36,40 @@ import shopping.order.proto.OrderResponse
...
       managementPortIndex: Int) {
+
+    private val serviceName = "shopping-cart-service"
+    private val config: Config =
+      DynamicTestConfig.endpointConfig(serviceName, grpcPort) // dynamic endpoints config
+        .withFallback(DynamicTestConfig.clusteringConfig(serviceName, managementPorts, managementPortIndex)) // dynamic cluster config
+        .withFallback(ConfigFactory.parseString(integrationSpecOverrides)) // test-specific overrides
+        .withFallback(ConfigFactory.parseResources("persistence-test.conf")) // persistence test overrides
+        .withFallback(ConfigFactory.load("application.conf")) // application config and reference config
+        .resolve()
```
ignasi35, line 67: This is the idiom I'm proposing.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
```diff
@@ -186,6 +198,7 @@ class IntegrationSpec
     // because responsible to close ScalikeJdbc connections
     testNode1.testKit.shutdownTestKit()
   }
+  // -------------------------------
```
ignasi35, line 201: In lines 100:201 I've reorganised the code and added separators and comments for readability. Maybe on a future refactor I'll introduce changes or extract code.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
```diff
@@ -0,0 +1,106 @@
...
+  /**
+   * Prepares config to (1) run multiple actor systems in a single JVM,
+   * (2) tune Akka Management to use the provided port,
+   * (3) set up Discovery to locate other nodes, and
+   * (4) wait cluster formation until all nodes are ready.
+   *
+   * @param managementPorts the management ports for all the nodes
+   * @param managementPortIndex the port to be used by _this_ node.
+   */
+  def forCluster(serviceName:String, managementPorts: Seq[Int], managementPortIndex: Int) = {
```
patriknw, line 34: explicit `: String =`

ignasi35 (commented):
comments for reviewers

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
```diff
@@ -0,0 +1,106 @@
...
+       """
+
+    s"""
+      ${forArtery()}
+      $singleJvmCluster
+      $management
+      $hardcodedDiscovery
+      $bootstrap
+      """
+  }
```
patriknw, line 90: this will be quite ugly in Java, but that's how it is

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
```diff
@@ -0,0 +1,106 @@
...
+         # unless a port is specified
+         port-name = "management"
+
+         # config service discovery never changes
+         stable-margin = 1 ms
+
+         # bootstrap without all the nodes being up
+         contact-with-all-contact-points = true
+       }
+       """
```
patriknw, line 81: this could move to a file in the Java version

patriknw (commented):
looking good

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
patriknw: My subjective opinion is that I don't like such comment blocks. They typically become stale or makes refactoring more difficult. I don't have a strong opinion if you find it important. Issue #52 might need more cleanup when we describe how to write tests.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
ignasi35: Yeah, let me see what I can do...

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/DynamicTestConfig.scala
ignasi35: You're right. In a previous iteration this had some dynamic content but I moved it into `val hardcodedDiscovery` above.

ignasi35:
I've pushed a first (quite ugly) version for the java implementation. I'll probably refactor to compose maps instead of impossible-to-read string concatenations.

octonato (commented):
> Integration tests must use random ports for gRPC, artery and akka management so those settings must be dynamic. On top of that, an integration test must start a 3-node (or 3 actorSystem) cluster while another needs only 1 node.

Would we simplify things if we just pick some fixed ports and hardcoded then? 

For instance, `integration-test1.conf`, `integration-test2.conf` and `integration-test3.conf`. Each would defined a fixed port for gRPC, remote, etc. 
Each actor system we use in the test would then load a different config with some shared imports as well.

This is exactly the approach we used for dev-mode with the different `localN.conf`. So, we could just take that same path.


The number of tests we have is not that big so a few config files isn't a big deal, IMO. 

I also think it will be much easier for users to understand what is going one.

ignasi35:
> Would we simplify things if we just pick some fixed ports and hardcoded then?

I suspect that could potentially explode. It would work on this sample app because we have one or two tests but in bigger codebases, with more extense test suites it may end up requiring too many .conf files to maintain.

octonato:
> I suspect that could potentially explode. It would work on this sample app because we have one or two tests but in bigger codebases, with more extense test suites it may end up requiring too many .conf files to maintain.

I don't think we should care about it. That's user responsibility to manage. They should judge when things are getting bigger and needs another approach. 

I think we get sometimes into that mindset because that's a sample, but also a kind of template for users to start a new project. 

When starting a new project, we expect it to grow and then the concern of exploding configuration files becomes more realistic. However, I think we should treat it first as a sample and we should not get in the trap of treating it as a template/framework.

Yes, some users will use it as a template, but I think as a developer they are entitled to arrange things the way they prefer. If we offer a code solution we kind of offer a half-baked framework for something that they can solve themselves.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
ignasi35: See https://github.com/akka/akka-platform-guide/pull/645 for an alternative

ignasi35:
Closed in favor of #645

PR #625, opened 2021-03-15
Improve CI reliability

andreaTP:
The CI tends to break on `05-shopping-cart-service` (both Scala and Java).
The problem is not reproducible locally, and, testing in isolation works as well.

Here we are splitting the tests into 2 groups, this makes the CI more resilient and might improve CI times ass well with very little copy-pasted configuration.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -48,7 +48,7 @@ jobs:
     name: Compile and Test Part 1 (Java)
     runs-on: ubuntu-18.04
     steps:
-      <<: &java_build
+      <<: *java_build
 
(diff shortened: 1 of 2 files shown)


ignasi35:
Noob question: why would splitting this job in two be more resilient? How does that work?

andreaTP:
> Noob question: why would splitting this job in two be more resilient? How does that work?

it makes the tests more "isolated", not claiming is a solution, but it might be a viable workaround.

andreaTP:
(this PR will require to mirror the changes into the repository settings)

patriknw (approved):
LGTM, strange that it wouldn't be able to run all same tests (sequentially) in one go but if this helps I'm fine with it

PR #629, opened 2021-03-16
Use Akka Platform BOM in examples

seglo:
References https://github.com/lightbend/akka-platform-meta/issues/149


akka-platform-bom/project/Dependencies.scala
```diff
@@ -0,0 +1,81 @@
+object Dependencies {
+  object Versions {
+    // FIXME: do we support 2.11 BOM's? some projects don't publish it anymore
```
patriknw, line 3: Do you mean Scala 2.11? No, we shouldn't support that.

I even think Scala 2.13 would be enough, and not even publish bom for 2.12, because I don't see a reason why a Java user would care to use the older 2.12.

akka-platform-bom/project/Dependencies.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    val AkkaCloudPlatform = "1.1.10"
```
patriknw, line 5: Not sure the version number should be synced with the operator version. I can see that both the operator and Akka libraries may change independent of each other. For example, we might want to publish a new bom to push out a security patch of akka-http.

akka-platform-bom/project/Dependencies.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  val akkaResilienceEnhancements = Seq(
+    "com.lightbend.akka" %% "akka-split-brain-resolver" % AkkaEnhancements,
+    // FIXME this is also published in the akka management org
+    //"com.lightbend.akka" %% "akka-lease-kubernetes" % AkkaResilienceEnhancements
+    "com.lightbend.akka" %% "akka-diagnostics" % AkkaEnhancements,
+    "com.lightbend.akka" %% "akka-fast-failover" % AkkaEnhancements
+  )
+
+// FIXME: wait for 3.0.0 (Alpakka), 2.1.0 (Alpakka Kafka), for Akka 2.6?
```
patriknw, line 74: will akka-stream-alpakka-cassandra be backwards compatible in a way that it can be used with akka-projection-cassandra and akka-persistence-cassandra?

Why wait? Wouldn't we just release a new bom when it's out?

akka-platform-bom/build.sbt
```diff
@@ -0,0 +1,10 @@
+import Dependencies._
+
+lazy val `akka-platform-bom` = Project(id = "akka-platform-bom", base = file("."))
+  .enablePlugins(BillOfMaterialsPlugin)
+  .settings(
+    name := "akka-platform-bom",
+    version := AkkaCloudPlatformVersion,
+    bomIncludeModules := Seq(akkaCoreBom, akkaHttpBom) ++ akkaManagement,
```
patriknw, line 8: nice that the other boms can be included

patriknw (commented):
looking good, it would be some work to keep it up to date but that's probably worth it

I think it's good to have it in this repo. That's the best we have for combining many things, and we can test it with the samples.

akka-platform-bom/build.sbt
ignasi35: We will need to test this. Details like the order in which the BOMs are declared in the user code or the depth in the Maven dependency tree can affect the version resolution.

See the paragraph on [`Note that using BOM's comes with its own set of problems...`](https://github.com/lightbend/akka-platform-meta/issues/96#issuecomment-732237090).

My point being, I think (but I'm not entirely sure without testing it) that this new bom should not _include_ the Akka and Akka-HTTP BOMs.

Se for example, the [spring bom](https://repo1.maven.org/maven2/org/springframework/spring-framework-bom/5.3.5/spring-framework-bom-5.3.5.pom) and the [spring boot dependencies](https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/2.4.3/spring-boot-dependencies-2.4.3.pom). Note how the `-dependencies` doesn't include the `-bom`.

ignasi35 (changes requested):

akka-platform-bom/project/Dependencies.scala
seglo: The example projects in the platform guide only target 2.13. If we're using them as a means to enforce compatibility maybe it would be safer to only publish this as 2.13 as well.

akka-platform-bom/project/Dependencies.scala
seglo: We discussed versioning the platform guide documentation itself (and it's examples, I guess). We could use that version.

akka-platform-bom/project/Dependencies.scala
seglo: I _think_ most of Alpakka `2.0.2` will work with Akka 2.6. In some cases, like Alpakka Cassandra we've already asserted it works in projections. Alpakka Kafka `2.0.7` should work as well for the same reason. I just wasn't sure if they should be included since those releases still compile against Akka 2.5, but I'll include them anyway.

akka-platform-bom/build.sbt
```diff
@@ -0,0 +1,10 @@
...
+    description := s"${description.value} (depending on Scala ${CrossVersion.binaryScalaVersion(scalaVersion.value)})"
+  )
```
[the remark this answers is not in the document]
seglo: > This new bom, OTOH, should not be called `-bom` but `-dependencies`. It _is_ formally a bom, but it lists dependencies from multiple organizations or versioning schemas. It is true that all the artifacts will (most likely) be under the Akka umbrella but, IMHO, we should add some (carefully selected) external projects. For example, I'd add Jackson artifacts on this new `akka-platform-dependencies` to lock the version better.

SGTM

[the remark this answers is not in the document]
seglo: > Off-topic: TIL there's a Jackson Bom which deals with the `jacksonVersion` vs `JacksonDatabindVersion` [mess](https://github.com/akka/akka/blob/master/project/Dependencies.scala?rgh-link-date=2021-03-17T12%3A04%3A54Z#L28-L29).
> 
> See [repo1.maven.org/maven2/com/fasterxml/jackson/jackson-bom](https://repo1.maven.org/maven2/com/fasterxml/jackson/jackson-bom/)

I'll look into adding that.

akka-platform-bom/project/Dependencies.scala
```diff
@@ -0,0 +1,81 @@
...
+    val AkkaHttp = "10.2.4"
+    val AkkaManagement = "1.0.10"
+    val AkkaProjections = "1.1.0"
+    val AkkaPersistenceCassandra = "1.0.5"
+    val AkkaPersistenceJdbc = "5.0.0"
+    val AkkaPersistenceCouchbase = "1.0"
+    val AkkaEnhancements = "1.1.13"
+    val Alpakka = "3.0.0"
+    val AlpakkaKafka = "2.1.0"
+  }
```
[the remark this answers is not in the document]
seglo: I missed that because it wasn't in the Lightbend build dependencies page. Maybe because it wasn't production ready for a time, but I think it is now.

akka-platform-bom/build.sbt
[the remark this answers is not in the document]
ignasi35: > I'll look into adding that.

Wait! I'm not sure having a BOM depend on other BOMs is possible and/or valid.
Maybe we should just _know_ that Jackson BOM exists and suggest users add it.

related to https://github.com/akka/akka-platform-guide/pull/629#discussion_r595953501

akka-platform-bom/build.sbt
seglo: Unfortunately it does not seem that you can include nested BOMs.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/ShoppingCartServer.java
```diff
@@ -7,7 +7,7 @@
 import akka.http.javadsl.ServerBinding;
 import akka.http.javadsl.model.HttpRequest;
 import akka.http.javadsl.model.HttpResponse;
-import akka.japi.Function;
+import akka.japi.function.Function;
```
seglo, line 10: I was getting a maven compilation failure over this..

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project shopping-cart-service: Compilation failure                                                      
[ERROR] /home/seglo/source/akka-platform-guide/docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/ShoppingCartServer.java:[25,40] incompatible types: akka.japi.fu
nction.Function<akka.http.javadsl.model.HttpRequest,java.util.concurrent.CompletionStage<akka.http.javadsl.model.HttpResponse>> cannot be converted to akka.japi.Function<akka.http.javadsl.model.HttpRequest,java.util.c
oncurrent.CompletionStage<akka.http.javadsl.model.HttpResponse>> 
```

seglo:
I'll move this to its own repository tomorrow. Should it live in `lightbend` since it references commercial things? `lightbend/akka-platform-dependencies`

patriknw:
> I'll move this to its own repository tomorrow. Should it live in lightbend since it references commercial things? lightbend/akka-platform-dependencies

Might be more future proof to have it at lightbend if we want to include other things outside akka/akka in the same bom (or manage several different boms in that same repo).

akka-platform-dependencies/project/Dependencies.scala
```diff
@@ -0,0 +1,193 @@
...
+    "com.typesafe.akka" %% "akka-persistence-testkit"    % Akka,
+    "com.typesafe.akka" %% "akka-persistence-typed"      % Akka,
+    "com.typesafe.akka" %% "akka-persistence-query"      % Akka,
+    "com.typesafe.akka" %% "akka-protobuf-v3"            % Akka,
+    "com.typesafe.akka" %% "akka-remote"                 % Akka,
+    "com.typesafe.akka" %% "akka-slf4j"                  % Akka,
+    "com.typesafe.akka" %% "akka-stream"                 % Akka,
+    "com.typesafe.akka" %% "akka-stream-testkit"         % Akka,
+    "com.typesafe.akka" %% "akka-stream-typed"           % Akka,
+    "com.typesafe.akka" %% "akka-testkit"                % Akka
```
patriknw, line 59: double check that it's the same list as the akka-bom

akka-platform-dependencies/project/Dependencies.scala
```diff
@@ -0,0 +1,193 @@
...
+    "com.lightbend.akka" %% "akka-persistence-multi-dc-testkit" % AkkaEnhancements
+  )
+
+  val akkaResilienceEnhancements = Seq(
+    "com.lightbend.akka" %% "akka-split-brain-resolver" % AkkaEnhancements,
+    // FIXME: this is also published in the akka management org. no longer commercial only?
+    //"com.lightbend.akka" %% "akka-lease-kubernetes" % AkkaEnhancements,
+    "com.lightbend.akka" %% "akka-diagnostics"   % AkkaEnhancements,
+    "com.lightbend.akka" %% "akka-fast-failover" % AkkaEnhancements
+  )
```
patriknw, line 133: I'd like some changes to enhancements, but will take the motivation offline.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/pom.xml
```diff
@@ -19,16 +19,19 @@
...
+        <akka-platform.version>1.0.0</akka-platform.version>
         <scala.binary.version>2.13</scala.binary.version>
+<!--
         <akka.version>2.6.13</akka.version>
         <akka-http.version>10.2.2</akka-http.version>
         <akka-management.version>1.0.9</akka-management.version>
+-->
         <!-- tag::akka-persistence-cassandra[] -->
-        <akka-persistence-cassandra.version>1.0.5</akka-persistence-cassandra.version>
+<!--        <akka-persistence-cassandra.version>1.0.5</akka-persistence-cassandra.version>-->
```
patriknw, line 30: cleanup before merge

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/src/main/java/shopping/cart/ShoppingCartServer.java
patriknw: could it be that it's using a different Akka HTTP or gRPC version now. Good to use latest anyway.

docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/pom.xml
seglo: I'll close this PR when I move to a new repo and update the example apps later. Just testing for now.

akka-platform-dependencies/project/Dependencies.scala
seglo: :+1:

seglo:
I moved the BOM to https://github.com/lightbend/akka-platform-dependencies/pull/1

I'll use this PR to update/validate the examples. We can merge it when the BOM is published.

akka-platform-bom/project/Dependencies.scala
seglo: Reset to 1.0.0

docs-source/docs/modules/how-to/examples/shopping-cart-service-java/pom.xml
```diff
@@ -228,7 +202,6 @@
             <plugin>
                 <groupId>com.lightbend.akka.grpc</groupId>
                 <artifactId>akka-grpc-maven-plugin</artifactId>
-                <version>${akka-grpc.version}</version>
```
ignasi35, line 231: Oh, wow! Controlling the grpc plugin version from the BOM is great!

docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/pom.xml
```diff
@@ -35,16 +29,9 @@
...
-                <groupId>com.typesafe.akka</groupId>
-                <artifactId>akka-bom_${scala.binary.version}</artifactId>
-                <version>${akka.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-            <dependency>
-                <groupId>com.typesafe.akka</groupId>
-                <artifactId>akka-http-bom_${scala.binary.version}</artifactId>
-                <version>${akka-http.version}</version>
```
ignasi35, line 47: I'm still not sure the `akka-platform-dependencies` should include the `akka-bom` and the `akka-http-bom`. 🤔

The resulting XML is a lot nicer, that's is true.

ignasi35 (commented):
A nice side effect of using BOms is that one can often shrink the list of explicit dependencies because the BOM will make sure that any transitive artifact will be loaded for the correct version. Any chance we can remove some stuff from the sample poms?

docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/pom.xml
seglo: I removed the project BOM's since they can't be nested in the way I expected.

seglo:
> A nice side effect of using BOms is that one can often shrink the list of explicit dependencies because the BOM will make sure that any transitive artifact will be loaded for the correct version.

I didn't know that. I'm not sure if that's something we want to do in this case since being explicit about what dependencies are required would be informative to the user.  It is _a lot_ of dependencies though, so I'm not opposed to the idea.

docs-source/docs/modules/microservices-tutorial/examples/00-shopping-cart-service-java/pom.xml
ignasi35: I'm not sure I understand what you mean.

I expected the `akka-platform-dependencies` to not set the Akka and Akka-HTTP versions and then my user code to depend on 3 BOMs:

```
<depMgmt>
  <akka-bom>
  <akka-http-bom>
  <akka-platform-dependencies>
</depMgmt>
```

The order of these three BOMs above is important. If my user code declared the following, the resolved artifacts would be different:

```
<depMgmt>
  <akka-platform-dependencies>
  <akka-bom>
  <akka-http-bom>
</depMgmt>
```

👆🏼 I think that in the second case whichever `akka-actor` version that's brought in transitively from any artifact in `akka-platform-dependencies` may have precedence over the version set up on `akka-bom`. Would have to test it though.

seglo: I think `akka-platform-dependencies` should include everything, otherwise you must deal with the same version compatibility issue, but at a higher level with multiple akka project BOMs.

ignasi35: Hmm, you're right. It's not uncommon for a BOM to depend on other BOMS:

https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-dependencies/2020.0.2/spring-cloud-dependencies-2020.0.2.pom

Let's keep it as you're doing.

seglo: I suppose we could include the BOM's and the dependencies, but it seems redundant.

ignasi35 (approved):

ignasi35:
`+90 −404` Nice!

PR #637, opened 2021-03-19
R in CRUD is for Read

ignasi35:


ignasi35:
Note to reviewers: I wouldn't even wait for the build.

Issue #588, opened 2021-02-19
failed: IntegrationSpec - Broker may not be available


patriknw:
Something with Kafka not available

https://github.com/akka/akka-platform-guide/pull/575/checks?check_run_id=1933884932

```
2021-02-19T08:43:15.5293126Z [2021-02-19 08:43:14,659] [INFO] [org.apache.kafka.clients.consumer.KafkaConsumer] [] [] [IntegrationSpec-akka.kafka.default-dispatcher-37] - [Consumer clientId=consumer-37a7844f-241c-4773-aa94-a9acc2b17d37-1, groupId=37a7844f-241c-4773-aa94-a9acc2b17d37] Subscribed to topic(s): shopping-cart-events
2021-02-19T08:43:15.5298316Z [2021-02-19 08:43:15,027] [WARN] [org.apache.kafka.clients.NetworkClient] [] [] [IntegrationSpec-akka.kafka.default-dispatcher-39] - [Consumer clientId=consumer-37a7844f-241c-4773-aa94-a9acc2b17d37-1, groupId=37a7844f-241c-4773-aa94-a9acc2b17d37] Connection to node -1 (localhost/127.0.0.1:9092) could not be established. Broker may not be available.
2021-02-19T08:43:15.5303105Z [2021-02-19 08:43:15,028] [WARN] [org.apache.kafka.clients.NetworkClient] [] [] [IntegrationSpec-akka.kafka.default-dispatcher-39] - [Consumer clientId=consumer-37a7844f-241c-4773-aa94-a9acc2b17d37-1, groupId=37a7844f-241c-4773-aa94-a9acc2b17d37] Bootstrap broker localhost:9092 (id: -1 rack: null) disconnected
2021-02-19T08:43:15.5306848Z [2021-02-19 08:43:15,092] [WARN] [org.apache.kafka.clients.NetworkClient] [] [] [kafka-producer-network-thread | producer-2] - [Producer clientId=producer-2] Connection to node -1 (localhost/127.0.0.1:9092) could not be established. Broker may not be available.
2021-02-19T08:43:15.5309617Z [2021-02-19 08:43:15,092] [WARN] [org.apache.kafka.clients.NetworkClient] [] [] [kafka-producer-network-thread | producer-2] - [Producer clientId=producer-2] Bootstrap broker localhost:9092 (id: -1 rack: null) disconnected
[...]

PR #639, opened 2021-03-19
Further split of the build + flaky test

ignasi35:
References #625

In #625, we moved the `05-` and the casssandra variants to separate job. This PR goes a bit further and also moves the final codebases to their own job. The resulting setup is:

 - `Final`: runs `shopping-cart`, `order` and `analytics` from the final-most codebase. This may be flaky in the `IntegrationSpec` on `shopping-cart`.
 - `Itermediate`: all the versions prefixed `01-` to `04-`. These are generally stable builds.
 - `Flaky`:  `05-` and `-cassandra` versions of `shopping-cart`.

I was tempted to _just_ move `shopping-cart`, `order` and `analytics` to `Flaky` and separate the jobs into `Flaky` vs `Non-Flaky`.

I'm open to other options.

### DIFF
--- a/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
+++ b/docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
@@ -215,7 +215,7 @@ class IntegrationSpec
 
       // first may take longer time
       val published1 =
-        kafkaTopicProbe.expectMessageType[proto.ItemAdded](20.seconds)
+        kafkaTopicProbe.expectMessageType[proto.ItemAdded](60.seconds)
       published1.cartId should ===("cart-1")
       published1.itemId should ===("foo")
       published1.quantity should ===(42)
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -23,8 +23,8 @@ jobs:
           scripts/copy-identical-files.sh
           git diff --exit-code --color || { echo "[error] Found modified file that was expected to be identical. Run scripts/copy-identical-files.sh"; false; }
 
-  java-compile-and-test-part1:
-    name: Compile and Test Part 1 (Java)
+  java-compile-and-test-part-final:
+    name: Compile and Test Final (Java)
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout
@@ -56,6 +56,30 @@ jobs:
       - name: "Shopping order service (Java)"
         run: scripts/mvn-test.sh docs-source/docs/modules/microservices-tutorial/examples/shopping-order-service-java
 
+  java-compile-and-test-part-intermediate:
+    name: Compile and Test Intermediate (Java)
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Checkout GitHub merge
+        if: github.event.pull_request
+        run: |-
+          git fetch origin pull/${{ github.event.pull_request.number }}/merge:scratch
+          git checkout scratch
+
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
+
+      - uses: actions/cache@v2.1.3
+        with:
+          path: ~/.m2/repository
+          key: maven-repo-${{ hashFiles('**/pom.xml') }}
+          restore-keys: maven-repo-
+
       - name: "00 Shopping analytics service (Java)"
         run: scripts/mvn-test.sh docs-source/docs/modules/microservices-tutorial/examples/00-shopping-analytics-service-java
 
@@ -77,8 +101,8 @@ jobs:
       - name: "04 Shopping cart service (Java)"
         run: scripts/mvn-test.sh docs-source/docs/modules/microservices-tutorial/examples/04-shopping-cart-service-java
 
-  java-compile-and-test-part2:
-    name: Compile and Test Part 2 (Java)
+  java-compile-and-test-flaky:
+    name: Compile and Test Flaky (Java)
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout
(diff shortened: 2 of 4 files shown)


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
```diff
@@ -215,7 +215,7 @@ class IntegrationSpec
 
       // first may take longer time
       val published1 =
-        kafkaTopicProbe.expectMessageType[proto.ItemAdded](20.seconds)
+        kafkaTopicProbe.expectMessageType[proto.ItemAdded](60.seconds)
```
andreaTP, line 218: Interestingly enough this is always the line where it breaks, the root cause is that Kafka has been somehow shut-down, should we start spawning one Kafka instance per test?
Does anyone guess what's the problem here?

andreaTP (approved):

patriknw (commented):
feels like we are trying things without understanding the root cause? is something spilling over from one part to another? some data remaining in Cassandra or Kafka? Then that would be a problem for users also if they repeatedly run the same test without restarting docker-compose. don't we create unique keyspace for the Cassandra stuff for each test run? do we have similar in place for Kafka topics and Postgres?

andreaTP:
@patriknw the problem is AFAIK that Kafka shut-down/stop to be operational, I don't have a clue on the cause, maybe we are consuming too much resources in CI?

ignasi35:
> feels like we are trying things without understanding the root cause? 

This is exactly what this PR is. We have a test implemented in several codebases that seems to be flaky. Either the test or the resources it depends on are failing. There are several possible causes and my first approach is to try to run the flaky versions of the test a bit more isolated.

If the isolation reduces flakiness then we can assert there are certain suspicions we had that no longer hold.

I agree 100% we still have not enough information re the actual cause.

ignasi35:
I currently work under the assumption that there are multiple (very unlikely) causes with similar symptoms.

Take, for example, [this failure](https://github.com/akka/akka-platform-guide/runs/2165004318). In this case, the `IntegrationTest` (the Java version) on `shopping-cart-java` run very early in the job (the job running tests first runs the final codebase, then runs 01, then 02, ...). 

So, in the failure mentioned above ([link](https://github.com/akka/akka-platform-guide/runs/2165004318)), it's unlikely that the database or the broker are offering left-over data from previous test executions. If that was the case, only `ItemOpuplarityIntegrationtest` has run before the failed test. I have run `ItemOpuplarityIntegrationtest` and `IntegrationTest` back to back multiple times over the same DB and broker instances locally hundreds of time without observing any issue.

In the failure mentioned above ([link](https://github.com/akka/akka-platform-guide/runs/2165004318)), OTOH, the VM is quite green. The failing test is the first one trying to read/write from Kafka. Either Kafka is not ready or it became available too late for the client testProbe to connect to it. It is important to note that this VM had just run `shopping-analytics-java` tests. That execution had all the docker containers in docker-compse to startup and then shutdown once tests were over. As a consequence, when `shopping-cart-java` arrives and tries to start all docker containers again, I wouldn't be surprised if the Kafka or ZooKeeper containers fail because the previous incarnation is not 100% gone. I thin I have seen similar issues on my local machine. 

There are other possible reasons.

docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala/src/test/scala/shopping/cart/IntegrationSpec.scala
ignasi35: This test is on `shopping-cart-{scala,java}`.  The jobs are organized to run in the following order:

1. start job
2. cd `shopping-analytics-xxx`
3. docker-compose up -d
4. sbt test (for `shopping-analytics-xxx`)
5. docker-compose stop
6. cd ..
7. cd `shopping-cart-xxx`
8. docker-compose up -d
9. sbt test (for `shopping-cart-xxx`)  <--- the failing tests is here
10. docker-compose stop
11. cd ..

I suspect (but have no proof) the problem is that steps `5.` and `8.` interfere causing the Kafka startup in `8.` to fail.

ignasi35:
I've just seen another failure where `05-shopping-cart-service-scala` was the first project running any compilation or tests on the VM of the job.

ignasi35:
Refs #588

ignasi35:
I think the main problem is we use a Kafka and a ZooKeeper images that are too slow to stop/start/stop/start. Our test suite will run something like the following (you can try that on your local machine):

```
echo "Starting analytics"
scripts/sbt-test.sh docs-source/docs/modules/microservices-tutorial/examples/shopping-analytics-service-scala 
echo "Completed Analytics"

echo "Starting shopping cart"
scripts/sbt-test.sh docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-scala
echo "done shopping cart"
```

This works on a local machine but during the `docker-compose` steps it's clear the Kafka and the ZooKeeper containers react very slowly (even on a powerful developer machine).

-----

I also noticed on a recent failure that after observing the `IntegrationSpec` failure, when the script proceeds to execute the `docker-compose stop` step the following output is logged:

```
...
[info] Suites: completed 4, aborted 0
[info] Tests: succeeded 11, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed tests:
[error] 	shopping.cart.IntegrationSpec
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 95 s (01:35), completed Mar 4, 2021, 6:46:26 PM
+ THE_EXIT_CODE=1
+ '[' -f docker-compose.yml ']'
+ docker-compose stop
Stopping shopping-cart-service-scala_postgres-db-test_1 ... 
Stopping shopping-cart-service-scala_postgres-db_1      ... 
Stopping shopping-cart-service-scala_postgres-db_1      ... done
Stopping shopping-cart-service-scala_postgres-db-test_1 ... done
+ cd /home/runner/work/akka-platform-guide/akka-platform-guide
+ exit 1
```

So it's clear the Kafka and the ZooKeeper containers were not started. See https://github.com/akka/akka-platform-guide/pull/611/checks?check_run_id=2033789315#step:7:1371
What's a bit annoying is that the counterpart `docker-compose up -d` did report both of those containers as started:

```
...
+ docker-compose up -d
Creating network "shopping-cart-service-scala_default" with the default driver
Creating shopping-cart-service-scala_postgres-db-test_1 ... 
Creating shopping-cart-service-scala_postgres-db_1      ... 
Creating shopping-cart-service-scala_kafka_1            ... 
Creating shopping-cart-service-scala_zookeeper_1        ... 
Creating shopping-cart-service-scala_postgres-db_1      ... done
Creating shopping-cart-service-scala_postgres-db-test_1 ... done
Creating shopping-cart-service-scala_zookeeper_1        ... done
Creating shopping-cart-service-scala_kafka_1            ... done
+ sbt 'test; scalafmtCheckAll'
[info] welcome to sbt 1.4.6 (AdoptOpenJDK Java 11.0.9)
...
```

See https://github.com/akka/akka-platform-guide/pull/611/checks?check_run_id=2033789315#step:7:11

andreaTP:
@ignasi35 should we close?

PR #645, opened 2021-03-22
Prefer conf files with hardcoded ports on tests

ignasi35:
This is an alternative to https://github.com/akka/akka-platform-guide/pull/624 following @octonato 's https://github.com/akka/akka-platform-guide/pull/624#pullrequestreview-617507603 on using hardcoded ports in tests and move all conf to files.

This is WIP and I only implemented the Java version (please ignore the changes in the scala codebase).

_UPDATE:_ I also closed #597 in favor of this PR. `ClusterBootstrap` and `AkkaManagement` will no longer need to be started programmatically and, instead, will use the config-based introduced in https://github.com/akka/akka-management/issues/850


docs-source/docs/modules/microservices-tutorial/examples/shopping-cart-service-java/src/test/java/shopping/cart/IntegrationTest.java
```diff
@@ -42,34 +30,38 @@
...
+        "local" + nodeId + "-test",
+        "management-3node-test",
+        "local-shared-test",
+        "cluster-test",
+        "grpc",
+        "serialization",
+        "persistence-test"
+    ).stream().map(confName ->
+        ConfigFactory.parseResources(confName + ".conf")
+    ).reduce((c1, c2) -> c1.withFallback(c2)).get();
```
ignasi35, line 61: This highlights the idiom I'm proposing.  Instead of using conf files with multiple settings, I opted towards using smaller conf files one can compose:

1. a conf file specific to the test I'm working where I can add all the necessary ovewrites for that test
2. a conf file with ports to be used by this node
3. a conf file tunning the discovery and bootstrap (may be a `1node` or a `3node`)
4. a conf file with values shared by all ndoes during tests
5. a conf file with overrides over `cluster.conf` to run a test
6. the default grpc conf, eventually we could have a grpc-test
7. same for serialization
8. a conf file with overrides over `persistence.conf` to run a test

This way conf files under the test/ scope are a lot smaller and easier to reason about. On each test, then, it's easier to pick the few conf files you need.

octonato: That code is doing what `include` does in typesafe config (more or less). 

The previous code was:

```java
private static Config sharedConfig() {
	return ConfigFactory.load("integration-test.conf");
}
```

why can't we simply have: 

```java
private static Config sharedConfig() {
	return ConfigFactory.load("integration-test-1.conf");
}
```

ignasi35: > (more or less)

The `more or less` part is the interesting part.

ignasi35: The main difference is that running the application locally makes the assumption that you run _everything_ while running a test only assumes you are in a test environment.

The test may want to reproduce a failure situation, or a simplified environment with 1 node or a mocked persistence layer. By limiting the usage of `include`'s in the `xxx-test.conf` files it's easier to reason about what settings are brought in. 

Keeping the composition of pieces in the test code itself or having one `test-name.conf` file which only does includes is a matter of test, I guess. Everything in code seems easier to maintain to me but I understand it depends on each case.

ignasi35:
We're putting this on hold until further notice.

PR #657, opened 2021-03-26
Fix the link for installing the AWS LB

andreaTP:
The AWS documentation includes the mandatory settings:
```
  --set serviceAccount.create=false \
  --set serviceAccount.name=aws-load-balancer-controller \
```

Otherwise the command:
```
helm upgrade -i aws-load-balancer-controller eks/aws-load-balancer-controller
```
will always fail.


ignasi35 (approved):
LGTM!

PR #673, opened 2021-03-30
Add AWS Load Balance Controller Helm chart link

ygree:
Resolves: #https://github.com/lightbend/akka-platform-meta/issues/165


andreaTP:
I agree with @thothmoses , we can add the github link but the AWS one contains detailed and relevant instructions.

thothmoses (approved):

PR #674, opened 2021-03-31
Use  IAM instead of ACM for importing certificate

ygree:
Resolves: #https://github.com/lightbend/akka-platform-meta/issues/167

### DIFF
--- a/docs-source/docs/modules/deployment/pages/aws-ingress.adoc
+++ b/docs-source/docs/modules/deployment/pages/aws-ingress.adoc
@@ -44,7 +44,9 @@ aws iam upload-server-certificate \
     --private-key file://MyKey.key
 ----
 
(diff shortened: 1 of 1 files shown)


andreaTP (approved):

andreaTP:
looks good! @ygree did you tested it?

ygree:
> looks good! @ygree did you tested it?

Yes. I could add, get and delete the cert. I didn't test the ingress though. Do you think it's worth doing so?

andreaTP:
> Do you think it's worth doing so?

yes, please 🙏  we have had so many hiccups there I would rather stay on the safe side and have another person go through the guide.

ygree:
> > Do you think it's worth doing so?
> 
> yes, please 🙏 we have had so many hiccups there I would rather stay on the safe side and have another person go through the guide.

I just finished testing and it all works as expected with `aws iam upload-server-certificate`.

PR #678, opened 2021-04-01
Revert "Add AWS Load Balance Controller Helm chart link"

andreaTP:
Reverts akka/akka-platform-guide#673

I noticed that the GH link doesn't mention the additional Helm settings that are required.
We can add the link as a reference, but shouldn't be the first go-to link.
Also, we should direct people in one direction, possibly.

### DIFF
--- a/docs-source/docs/modules/deployment/pages/aws-ingress.adoc
+++ b/docs-source/docs/modules/deployment/pages/aws-ingress.adoc
@@ -7,7 +7,7 @@ To expose your application's gRPC or HTTPS endpoints publicly to the internet yo
 
 == Install the LoadBalancer Controller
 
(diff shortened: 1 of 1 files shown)


andreaTP:
cc. @ygree

ygree:
References: #https://github.com/lightbend/akka-platform-meta/issues/165

andreaTP:
@ygree the helm installation steps will fail if you don't provide:
```
  --set serviceAccount.create=false \
  --set serviceAccount.name=aws-load-balancer-controller \
  ```

thothmoses (approved):
the GH repo is actually mentioned in the AWS article. If you decide to retain the second link you can append it with "Additional information and resources are in https://github.com/aws/eks-charts/tree/master/stable/aws-load-balancer-controller[AWS Load Balancer Controller Helm chart for Kubernetes Github Repo {tab-icon}, window="tab"]

andreaTP:
@thothmoses I like this approach, do you wanna follow up?

thothmoses:
Ok, I have added the additional info.

PR #687, opened 2021-04-15
Updated some graphics with reflow and conversion to SVG

thothmoses:
References #xxxx


docs-source/docs/modules/deployment/pages/cloud-deployment.adoc
```diff
@@ -13,7 +13,7 @@ The following image shows the components required to deploy an Akka Microservice
 * Creating and publishing the Docker image
 * Updating the deployment descriptor and applying it to the Kubernetes cluster
 
-image::deployment-overview1.png[Deployment 1]
+image::deployment-overview1.SVG[Deployment 1]
```
patriknw, line 16: below callout descriptions should be adjusted since the numbers in the picture changed

patriknw (commented):
nice, are there some original source files for the images that you can add to https://github.com/akka/akka-platform-guide/tree/main/docs-source/resources/ ?

docs-source/docs/modules/deployment/pages/cloud-deployment.adoc
thothmoses: The numbers didn't change - I just re-flowed the diagram to remove the cross-overs.

thothmoses: All of the source files are publicly available from Docker and K8s. The rest of the diagrams are in XAR files for XARA. I can generate AI files and attach them if you think that would be useful. But most Vector applications can edit the SVG directly.

patriknw: I must have been looking at overview2. All good with the numbers.

I still don't know how I would edit these pictures without asking you, but maybe that's ok?
My experience of raw SVG is that it's not enough for reproducing the original look of the pictures if using a different editor.

patriknw (approved):
LGTM, but I think we will get into future problems of editing the pictures if we only have raw svg.

PR #699, opened 2021-05-03
URL validation rule updates

seglo:
The link validator was silently failing. There were some violations, but the github action didn't return an error.

I have an [upstream fix for the validator](https://github.com/ennru/site-link-validator/pull/3), but this PR fixes the violations in the platform guide in the meantime. These updates are not critical, no URLs are actually broken.


seglo:
I tested that `link-validator:0.2.1` works as expected in the build.

andreaTP (approved):

PR #703, opened 2021-05-05
Improve GCP Helm to take into account publishing delays

andreaTP:
Publishing to GCP takes up to days, this PR address installation/upgrades during that time.

### DIFF
--- a/docs-source/docs/modules/deployment/pages/gcp-install.adoc
+++ b/docs-source/docs/modules/deployment/pages/gcp-install.adoc
@@ -87,11 +87,11 @@ kubectl apply -f "https://raw.githubusercontent.com/GoogleCloudPlatform/marketpl
 
 . Update the Akka Microservices CRD to latest:
 +
-[source,shell script]
+[source,shell script,subs="attributes+"]
 ----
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/deployment/pages/gcp-install.adoc
```diff
@@ -83,12 +83,15 @@ secret/akka-cloud-platform-1-license created
...
 ----
 
+. Verify the latest version already published to GCP https://console.cloud.google.com/gcr/images/cloud-marketplace/GLOBAL/lightbend-public/akka-cloud-platform?gcrImageListsize=30[here] and take note of the it.
+
 . Update the Akka Microservices CRD to latest:
 +
 [source,shell script]
 ----
-kubectl apply -f  https://lightbend.github.io/akka-operator-helm/akka-operator/crds/v1/akka-microservices-crd.yml
+kubectl apply -f  https://raw.githubusercontent.com/lightbend/akka-operator-helm/v1.1.19/akka-operator/crds/v1/akka-microservices-crd.yml # <1>
```
patriknw, line 92: If we have to include the version we should at least make that a variable so that we can change it at one place.

andreaTP: In this case is just "indicative", but I agree, implemented using an asciidoc `attribute`.

patriknw (approved):
LGTM

Issue #705, opened 2021-05-06
in How to use Cassandra instead of PostgreSQL section, full download link comes back with 404


dantodor:
I think the bug description is clear enough.

ignasi35:
Thanks for the heads up!
I've confirmed that _both_ java and scala links in https://developer.lightbend.com/docs/akka-platform-guide/how-to/cassandra-alternative.html#_source_downloads lead to 404. We'll look into it.

ignasi35:
I suspect this was a one-time error on the deployment script that takes this guide and publishes it to dev.lb.com. I'll retrigger the build.

thothmoses:
@ignasi35 For some reason these are still broken on the live doc - I tried it locally and there was no problem - but looking at the online repo it looks like the 'attachments' folder is being stripped out - present locally but not on Main or Current. IIt is listed in the gitignore file  with a note that implies these zip files are generated? Some guidance, please.

thothmoses:
@adriaanm Since Ignasi is out could you take a look at this?

patriknw:
I agree that this is important to look into. Could it even be that the other sample zip files are outdated (not updated any more) and we serve something old from the doc site?
cc @RayRoestenburg

RayRoestenburg:
I'll have a look. worked locally for me as well, on mac you need to use gnu-sed (https://gist.github.com/andre3k1/e3a1a7133fded5de5a9ee99c87c6fa0d#gistcomment-3524266).

RayRoestenburg:
I think I see what is happening, the zips are ending up in the wrong directory in the published tar ball:
<img width="561" alt="image" src="https://user-images.githubusercontent.com/156425/121562718-01050a00-ca1a-11eb-85d7-d5cd8b5ad9a6.png">
@ignasi35 @patriknw

thothmoses:
@patriknw @RayRoestenburg Yep, it should be under how 'how-to' How do we fix that?

PR #708, opened 2021-05-07
Fail script in case of intermediate error

ignasi35:
References #705

The script I'm modifying is not cross-platform, it uses some linux-especific syntax for `sed` and I haven't tested it yet. 

Apart from the platform-specific syntax, it's possible this script failed silently in some step before if/when a replace operation didn't find anything to replace of the source/target folders where incorrect. These silent errors used to be fine but something failed in the latest deployment (2021-05-06) of Techhub and the `prepare-downloads or some other step of the publishing of the site failed silently leaving (at least) a couple of broken links.

### DIFF
--- a/scripts/prepare-downloads.sh
+++ b/scripts/prepare-downloads.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 declare -r tutorial_sources="${PWD}/docs-source/docs/modules/microservices-tutorial/examples"
 declare -r tutorial_attachments="${PWD}/docs-source/docs/modules/microservices-tutorial/assets/attachments"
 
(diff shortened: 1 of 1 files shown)


ignasi35:
I stand corrected, there is output available for the build. I just didn't know what to look for:

https://ci.lightbend.com/job/techhub-publisher/label=techhub-ci/10285/consoleFull

Still, there's no error on that build.

scripts/prepare-downloads.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
ignasi35, line 3: Hmm, failing is not a good idea. 

It's better to publish a half-baked site than fail the whole publication.

ignasi35:
```suggestion
```

PR #711, opened 2021-05-07
Remove unused volumes in kafka service

ihostage:
Looks like these mappings were copied from an [example](https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml). 
But I think it's not needed really.


ihostage:
Possibly It can be useful for `dind` on CI for forwarding docker socket from a parent env for images that building by docker-compose (instead of built images from Docker Hub). But it's not our case. I haven't other ideas. 🤷‍♂️

ihostage:
Hi @octonato! 👋 
How about merge that?

PR #604, opened 2021-03-02
Update create_user_tables.sql

staniopolis:
We do not use "version" field in ItemPopularity Projection. It case an server Error because it can't be null

References #xxxx

PR #717, opened 2021-05-15
Fix wrong create_user_tables.sql reference

martijnblankestijn:
The scala solution does not have the version column in its data access.

Issue #604 was closed, but mentions the same issue.


patriknw (approved):
LGTM, thanks

ennru (approved):
LGTM.

PR #726, opened 2021-06-11
Fix attachements location for the how-to guide

### LINKED ISSUE #705: in How to use Cassandra instead of PostgreSQL section, full download link comes back with 404

andreaTP:
Should fix #705


RayRoestenburg (approved):
LGTM

PR #742, opened 2021-07-15
Pulumi: document how to install Telemetry backends

marcospereira:
## Status

It can be merged at any time, but DO NOT merge it into the `current` branch before emerging lightbend/akka-cloud-platform-deploy#12.

## Changes

- aws: Update quickstart guide to include new Pulumi changes
- gcp: Update quickstart guide to include new Pulumi changes

Reference: lightbend/akka-cloud-platform-deploy#12.

PR #743, opened 2021-07-16
Fix a few minor issues

marcospereira:
This is a follow up to #742 to fix some minor issues as suggested by @thothmose.﻿


thothmoses (approved):
Good to go with those changes

PR #759, opened 2021-08-11
Wait for Cassandra server to start before running the tests

marcospereira:
The latest version of Cassandra docker image uses Cassandra 4. This version takes longer
to start the service, possibly due to this issue:

https://issues.apache.org/jira/browse/CASSANDRA-15850

That is causing some builds to fail since the Cassandra server may no be up when the
tests start. The fix then is to have a healthcheck on docker compose and only run
the tests after the Cassandra server is healthy.

### DIFF
--- a/docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/docker-compose.yml
+++ b/docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/docker-compose.yml
@@ -1,9 +1,23 @@
 version: '2.2'
 services:
-  cassandra:
+  cassandra-service:
     image: cassandra:latest
     ports:
       - "9042:9042"
+    healthcheck:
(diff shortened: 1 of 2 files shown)


docs-source/docs/modules/how-to/examples/shopping-cart-service-cassandra-java/docker-compose.yml
```diff
@@ -1,9 +1,23 @@
 version: '2.2'
 services:
-  cassandra:
+  cassandra-service:
```
debasishg, line 3: do we need the postfix `service` since it's already under `services` ?

patriknw: there is still a `cassandra:` below

patriknw (approved):
LGTM

PR #793, opened 2022-09-07
Update footer, remove Platform reference from Ui elements

rasummer:
@ennru this updates the Akka Platform Guide with the new footer. I've aligned the drop down menus so that they reflect the footer changes. Also I've updated the logo so it's just Akka. Basically the visual aspects and UI elements. 

The actual content of the guide still uses Akka Platform in places,  I will let you guys update the actual docs as it will likely need some word smithing :)

Akka-Platform-Guide-Akka-Guide (1)

Akka-Platform-Guide-Akka-Guide


ennru (approved):
LGTM.

PR #794, opened 2022-09-07
Change license

johanandren:
https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka


patriknw:
let's update the dependencies first

patriknw:
No need to change here

PR #800, opened 2022-10-13
Remove Akka Cloud Platform from the guide

patriknw:


patriknw:
We should also add a redirect for the removed `deployment/*` pages. Is that done in some Antora config?

patriknw:
Antora page-aliases is not intended for redirect of whole directory tree like this
https://docs.antora.org/antora/latest/page/page-aliases/#page-aliases-and-bulk-url-redirects

Maybe @JustinPihony can help with adding a .htaccess redirect for https://developer.lightbend.com/docs/akka-platform-guide/deployment/* to https://developer.lightbend.com/docs/akka-platform-guide/index.html

PR #806, opened 2022-10-26
Publish latest

patriknw:
DO NOT SQUASH


patriknw:
The problem is that `.travis/known_hosts` was removed. I'll fix.

PR #827, opened 2023-03-22
Re-add Akka Platform AWS deployment pages

ennru:
Reverts large parts of 
- #800


sebastian-alfers:
I wonder if that will even work since we also renamed the repo from `akka-patform-guide` to `akka-guide`.

sebastian-alfers:
maybe we have to keep the redirects in place and just deploy this PR?

sebastian-alfers (approved):
LGTM

PR #836, opened 2023-04-17
bump: dependencies for 22.10.5

johanandren:


johanandren:
Completely confused how we should deal with dependency bumps here, if that wasn't already obvious.

PR #837, opened 2023-04-19
bump: Akka Management 1.3.0

johanandren:


johanandren:
I should add this to the other bump PR instead

